Issue #59, opened 2022-12-29
Linux commands `date` and `ls` behave differently on Mac


jsolbrig:
# Requested Update

### Description
The standard `date` and `ls` commands behave differently on Mac. Code that uses them should be updated to work correctly on Mac.

### Background and Motivation
The following scripts need to be updated to work correctly on Mac:
- `date` command
    - setup.sh
    - tests/download_noaa_aws.sh
- `ls` command
    - ./geoips/interface_modules/output_formats/text_winds.py

For scripts that use `date`, it should be updated to use `gdate` on Mac and raise an error if `gdate` is not found.

For the script that uses `ls`, since it is a Python script and the `ls` is just used to retrieve a file creation time for use in a log statement, it should be updated to use `os.stat` instead.

### Code to demonstrate issue
In scripts that use `date`, this should be added to the top of the script:
```bash
date_cmd=date
if [[ $OSTYPE == 'darwin'* ]]; then
    date_cmd="$(which gdate)"
    if [[ $? -ne 0 ]]; then
        echo "On Mac, please install gdate. For example, brew install coreutils."
        exit 1
    fi  
fi
```
and all uses of `date` should be replaced with `$date_cmd`.

In the script that uses `ls`, calls to `ls --full-time` should be replaced with `datetime.fromtimestamp(os.stat(fname).st_ctime)`.

### Checklist for Completion
* [x] Fix setup.sh
* [x] Fix tests/download_noaa_aws.sh
* [x] Fix geoips/interface_modules/output_formats/text_winds.py

PR #60, opened 2022-12-29
Linux commands date and ls behave differently on mac

jsolbrig:
# Related Issues
fixes NRLMMD-GEOIPS/geoips#59

# Testing Instructions
These changes have been tested on Mac. They should also be tested on Linux.
- [ ] Run `setup.sh` and ensure there are no errors.
- [ ] Ensure that you have `test_data_smap` installed and run `./tests/scripts/smap.unsectored.text_winds.sh` to ensure that there are no errors.

# Summary
### Fix linux commands `date` and `ls` on Mac
* Updated `setup.sh` and `tests/download_noaa_aws.sh` to use `gdate` on Mac
* Updated `geoips/interface_modules/output_formats/text_winds.py` to use `os.stat` rather
  than `ls --full-time` to get file creation time.


mindyls (approved):
I will test this during the interface module tests - these are isolateed enough that it should be obvious if they break.

Issue #92, opened 2023-02-07
Replace FilenameFormattersInterface class name with FilenameFormats


mindyls:
# Requested Update

### Description
* Replace FilenameFormattersInterface class name with FilenameFormats

### Background and Motivation
* Typo due to mid-PR update of class names.

### Environment
* geoips >= 1.6.0

### Code to demonstrate issue
* test_full_install.sh works.

### Checklist for Completion
* [ ] Update class name in filename_formats.py https://github.com/NRLMMD-GEOIPS/geoips/blob/main/geoips/interfaces/filename_formats.py

Issue #56, opened 2022-12-29
Develop an updated method of comparing annotated imagery


jsolbrig:
# Requested Update

### Description
Develop an updated method of comparing annotated imagery.

### Background and Motivation
The current method of testing annotated imagery uses ImageMagick's `compare` command. The command used looks like this:
```
compare -verbose -quiet -metric ae -fuzz 5% <control image> <test image> <diff output image>
```

This test does not appear to work as expected. It is still comparing individual pixels but with an allowed variation of +/-5% in each color. This is not adequate for comparing annotated imagery because we need to allow for shifts in the location of different elements of the imagery as shown in this difference image:
![diff_test_output_20200518_073601_IO012020_amsr2_gcom-w1_89H-Physical_140kts_100p00_res1p0-cr300](https://user-images.githubusercontent.com/1684491/209893223-809179f2-6458-49eb-95bb-b7107794b6e9.png)

### Alternative Solutions
Simply stop testing annotated imagery. This could work temporarily but would cause us to miss issues that arise with annotated imagery.

jsolbrig:
I think that the direction we should go with this is to start testing directly in Python rather than using Imagemagick. That should give us a lot more flexibility in how we perform our tests. I wonder if we could pickle our matplotlib objects and compare their contents.

PR #93, opened 2023-02-09
Rename FilenameFormattersInterface class to FilenameFormats

lance-wilson:
# Reviewer Checklist

# Related Issues
fixes NRLMMD-GEOIPS/geoips#92

# Testing Instructions

None

# Summary

### Bug fixes
* filename_formats.py:
  * Renamed FilenameFormattersInterface class to FilenameFormats

# Output

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@
     # # # for more details. If you did not receive the license, for more information see:
     # # # https://github.com/U-S-NRL-Marine-Meteorology-Division/
 
+## GEOIPS/geoips#92: 2023-02-09, update FilenameFormats class name
+### Bug fixes
+* filename_formats.py:
+  * Renamed FilenameFormattersInterface class to FilenameFormats
+
 ## NRLMMD-GEOIPS/geoips#69, 2023-02-06, bug fixes
 ### Bug fixes
 * Remove extra "," at the end of filename formats without covg list
--- a/geoips/interfaces/filename_formats.py
+++ b/geoips/interfaces/filename_formats.py
@@ -1,7 +1,7 @@
 from geoips.interfaces.base import BaseInterface, BasePlugin
 
 
-class FilenameFormattersInterface(BaseInterface):
+class FilenameFormats(BaseInterface):
     name = "filename_formats"
     entry_point_group = "filename_formats"
     deprecated_family_attr = "filename_type"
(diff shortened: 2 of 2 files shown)


jsolbrig (commented):
This looks good to me. @lance-wilson Can you please try running a few of the tests to be sure that the change didn't impact anything? I don't think it should, but it's best to test. I'd recommend testing at least one product that uses the single_source procflow and one that uses the config_based procflow.

lance-wilson:
@jsolbrig I ran the test_full_install.sh and it looks like all of the output was generated fine but the comparisons failed (usually the text files, I think; the image compares seemed to all work).  My fork wasn't setup with conda and my directory structure doesn't quite match the expectation, so I'm not sure if that is messing something up.  Sample images below.
Output image of abi.config_based_output_low_memory.sh:
20200918_195020_AL202020_abi_goes-16_Infrared_110kts_100p00_1p0
Output of abi.static.Visible.imagery_annotated.sh:
20200918 195020 goes-16 abi Visible goes16 41p12 noaa 10p0

jsolbrig:
@lance-wilson I think that all looks correct. Can you take a look at the output comparison images to see what they look like? I would expect all of the clean imagery to pass the comparison tests, but anything annotated may not. You can find the image diffs in `geoips/tests/outputs/<test_name>/diff_test_output_*/diff_test_output_*.png`. What I would expect to see is that the differences (marked in red) are all related to annotations like text and gridlines.

lance-wilson:
I think these look fine but I'm not sure they are rendering correctly.
diff_test_output_exact_20200918 195020 goes-16 abi Visible goes16 41p12 noaa 10p0
diff_test_output_exact_20200918_195020_AL202020_abi_goes-16_Infrared_110kts_100p00_1p0

mindyls (approved):
I'm convinced!  Good to go from my perspective.

jsolbrig:
Actually, those look exactly as I'd expect them to! Basically, those images appear to be faded because they are showing a diff of your output image against a pre-generated comparison image. The regions where they match are light. The regions where they don't match are marked in bright red. You can see a few pixels here and there, mainly in the text, that are bright red. That is where the images don't quite match.

Unfortunately, until we handle #56, we're going to keep seeing this kind of issue where our output doesn't quite match the comparison images. When they look like what you got, though, things are okay and working as correctly as they can.

Issue #117, opened 2023-04-03
Convert YAML files for `boundaries` interface


jsolbrig:
# Requested Update

### Description
*  [ In 2-3 sentences briefly describe the required bug fix or feature with this issue. ]

### Background and Motivation
*  [ OPTIONAL: Provide additional detailed information and background related to this issue ]

### Alternative Solutions
*  [ OPTIONAL: If appropriate, list alternative fixes that could be considered]
  
### Environment
*  [ IF APPROPRIATE: Corresponding repository versions ]

### Code to demonstrate issue
*  [ IF APPROPRIATE: Can include steps here to identify/demonstrate bug fix or desired feature ]

### Checklist for Completion
* [ ]

jsolbrig:
@evrose54 

Go ahead and create a PR for your work. You can continue pushing changes to your branch and they will be reflected in the PR.

To test your schema, do the following:

### Get my changes into your branch
```
git fetch
git checkout integration
git pull
git checkout <your branch name>
git merge integration
```
Once you’ve done that, you should run
```
pip uninstall geoips
pip install -e .[dev,lint]
```
Lastly, edit `geoips/schema/__init__.py` to change `from plugin_validator import validate` to `from .plugin_validator import validate`.

### Testing
In Python:
```python
> from geoips.schema import validate
> validate("<path to a plugin file>")
```

PR #128, opened 2023-04-04
108 develop schema for boundaries interface plugins

jsolbrig:
# Reviewer Checklist

# Related Issues
fixes NRLMMD-GEOIPS/geoips#[108]

# Testing Instructions
Compare the fields in `geoips/schema/boundaries.yaml` against the fields in the old boundaries yaml files located in `geoips/plotting_params/boundaries`. Ensure that the schema captures everything and uses appropriate types.

Real testing will occur in #117.

# Summary
Adds a schema for boundaries plugins.

### DIFF
--- a/docs/source/releases/v1_10_0_dev0.rst
+++ b/docs/source/releases/v1_10_0_dev0.rst
@@ -1,0 +1,12 @@
+Version 1.10.0.dev0 (2023-04-04)
+********************************
+
+Enhancements
+============
+Add a schema for validating `Boundaries` plugins.
+-------------------------------------------------
+* From issue NRLMMD-GEOIPS/geoips#108: 2023-04-04, Add a schema for `Boundaries` plugins.
+
+::
+
+    modified: /geoips/schema/boundaries.yaml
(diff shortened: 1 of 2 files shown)


mindyls (approved):
I just want to confirm this is a reasonable thing to do (setting gridlines / map boundaries options via a YAML config file).

https://github.com/NRLMMD-GEOIPS/geoips/blob/696c8c98cc4b80d3f3a5b3980036252803cd7c20/geoips/image_utils/maps.py#L593

https://github.com/NRLMMD-GEOIPS/geoips/blob/696c8c98cc4b80d3f3a5b3980036252803cd7c20/geoips/image_utils/maps.py#L693

cartopy does have interesting ways of handling these types of annotations, so maybe it is worth having a "simplified" upstream method for setting these parameters?

Also note maps.py will need updated to handle the new format.

Output formats can roll their own plotting, and not use these utilities at all, but is it worth having here?

If so, this looks good to me as a replacement for the original boundaries/gridlines YAMLs:

https://github.com/NRLMMD-GEOIPS/geoips/blob/integration/geoips/yaml_configs/plotting_params/boundaries/tc_pmw.yaml

https://github.com/NRLMMD-GEOIPS/geoips/blob/integration/geoips/yaml_configs/plotting_params/gridlines/tc_pmw.yaml

PR #134, opened 2023-04-04
Fix version import in interfaces/base.py

jsolbrig:
# Reviewer Checklist

# Related Issues
fixes NRLMMD-GEOIPS/geoips#[133]

# Testing Instructions

Try the following:
```
> python
> import geoips
> print(geoips.__version__)
> print(geoips.interfaces.bases.geoips_version)
```
Both should output a valid version string.

# Summary

This ended up being a little more complicated than I thought. There were two version files, `geoips/version.py` and `geoips/__version__.py`.

To fix this, I removed `geoips/__version__.py` in favor of the file that is auto-generated by setuptools. I then updated `geoips/__init__.py` and `geoips/interfaces/base.py` to correctly use `geoips/version.py`.

# Output


jsolbrig:
The issue that lead to this was incorrect. Closing.

PR #185, opened 2023-04-25
Updating release notes to support 1.10.0a0 release

mindyls:
I accidentally started all the plugins reorg PRs with 1.10.0a0. Making this one 1.10.0a0, and will increment the others BEFORE merging (no other changes besides moving the release notes for each approved 1.10.0aN PR)

Note 1.10.0.dev0 DOES NOT work with setuptools_scm, thus the switch to 1.10.0a0.  Unclear exactly what will and will not work, but I know this does, and it is concise.

### DIFF
--- a/docs/source/releases/index.rst
+++ b/docs/source/releases/index.rst
@@ -22,7 +22,7 @@ Version 1.10
 .. toctree::
    :maxdepth: 1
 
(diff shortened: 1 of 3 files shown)


jsolbrig (approved):
Looks good to me!

PR #214, opened 2023-05-01
Removing all old-style YAML configs

mindyls:
Intermediate step towards updating to use all new interfaces.


mindyls:
Splitting up into PR for each interface.

PR #317, opened 2023-09-06
56 compare images with python and mpl

evrose54:
# Related Issues
fixes NRLMMD-GEOIPS/geoips #56 

# Testing Instructions
To test these changes, I've been using the two files listed below to compare annotated imagery. However, any test script which compares imagery will be fine for this purpose. They pass without any problems.

  - geoips/tests/scripts/abi.static.Infrared.imagery_annotated.sh
  - geoips/tests/scripts/amsr2.tc.89H-Physical.imagery_annotated.sh

# Summary
Previously, the functionality in GeoIPS to compare imagery used 'imagemagick compare' methods. This would work for all clean imagery, however sometimes failed for annotated imagery. To circumvent this problem and allow for more flexibility on our side, we've implemented comparisons using PIL and Numpy, which result in the same figures, without having an extra dependency. While it takes a slight amount longer, this should be beneficial in the long run.

  - modified: geoips/geoips/compare_outputs.py

# Output
The output shown below was captured after running 'geoips/tests/scripts/amsr2.tc.89H-Physical.imagery_annotated.sh'.

Screenshot 2023-09-06 at 12 41 25 PM

diff_test_output_exact_20200518_073601_IO012020_amsr2_gcom-w1_89H-Physical_140kts_100p00_res1p0-cr300

### DIFF
--- a/geoips/compare_outputs.py
+++ b/geoips/compare_outputs.py
@@ -214,37 +214,55 @@ def images_match(output_product, compare_product, fuzz="5%"):
     bool
         Return True if images match, False if they differ
     """
-    out_diffimg = get_out_diff_fname(compare_product, output_product)
-    exact_out_diffimg = get_out_diff_fname(
-        compare_product, output_product, flag="exact_"
-    )
+    # out_diffimg = get_out_diff_fname(compare_product, output_product)
+    # exact_out_diffimg = get_out_diff_fname(
+    #     compare_product, output_product, flag="exact_"
+    # )
 
-    call_list = [
-        "compare",
-        "-verbose",
-        "-quiet",
-        "-metric",
-        "ae",
-        "-fuzz",
-        fuzz,
-        output_product,
-        compare_product,
-        out_diffimg,
-    ]
+    # call_list = [
+    #     "compare",
+    #     "-verbose",
+    #     "-quiet",
+    #     "-metric",
+    #     "ae",
+    #     "-fuzz",
+    #     fuzz,
+    #     output_product,
+    #     compare_product,
+    #     out_diffimg,
+    # ]
 
-    exact_call_list = [
-        "compare",
-        "-verbose",
-        "-quiet",
-        "-metric",
-        "ae",
-        output_product,
-        compare_product,
-        exact_out_diffimg,
-    ]
+    # exact_call_list = [
+    #     "compare",
+    #     "-verbose",
+    #     "-quiet",
+    #     "-metric",
+    #     "ae",
+    #     output_product,
+    #     compare_product,
+    #     exact_out_diffimg,
+    # ]
+    import numpy as np
+    from PIL import Image
+    from matplotlib import pyplot as plt
 
-    LOG.info("**Running %s", " ".join(call_list))
-    fullimg_retval = subprocess.call(call_list)
+    LOG.info("**Comparing output_product vs. compare product")
+    comp_img = np.asarray(Image.open(output_product))
+    out_img = np.asarray(Image.open(compare_product))
+    diff_images = comp_img - out_img
+    fullimg_retval = 0 if np.all(diff_images == 0) else 1
+    fig = plt.figure()
+    fig_shape = np.shape(diff_images)
+    fig.set_size_inches(fig_shape[2], fig_shape[1])
+    left_margin = 0.0
+    right_margin = 1.0
+    bottom_margin = 0.0
+    top_margin = 1.0
+    main_ax = plt.Axes(fig,
+                       [left_margin, bottom_margin,
+                        right_margin - left_margin,
+                        top_margin - bottom_margin,],)
+    # fullimg_retval = subprocess.call(call_list)
     LOG.info("**Done running compare")
 
     # call_list = ['compare', '-verbose', '-quiet',
@@ -275,14 +293,8 @@ def images_match(output_product, compare_product, fuzz="5%"):
         LOG.info("    *** BAD Images do NOT match exactly ***")
         LOG.info("    ***   output_product: %s ***", output_product)
         LOG.info("    ***   compare_product: %s ***", compare_product)
-        LOG.info("    ***   out_diffimg: %s ***", out_diffimg)
-        LOG.info("    ***   exact_out_diffimg: %s ***", exact_out_diffimg)
         LOG.info("    ***************************************")
         return False
-
-    LOG.info("**Running exact %s", " ".join(exact_call_list))
-    fullimg_retval = subprocess.call(exact_call_list)
-    LOG.info("**Done running exact compare")
 
     if fullimg_retval != 0:
         LOG.info("    ******************************************")
@@ -293,9 +305,9 @@ def images_match(output_product, compare_product, fuzz="5%"):
         LOG.info("    *** GOOD Images match exactly ***")
         LOG.info("    *********************************")
     # Remove the image if they matched so we don't have extra stuff to sort through.
-    from os import unlink as osunlink
+    # from os import unlink as osunlink
 
-    osunlink(out_diffimg)
+    # osunlink(out_diffimg)
     return True
 
 
--- a/geoips/compare_outputs.py
+++ b/geoips/compare_outputs.py
@@ -226,7 +226,7 @@ def images_match(output_product, compare_product, fuzz="5%"):
     out_img = Image.open(output_product)
     comp_img = Image.open(compare_product)
     diff_img = Image.new(mode="RGB", size=comp_img.size)
-    diff_arr = np.asarray(comp_img) - np.asarray(out_img)
+    diff_arr = np.abs(np.array(comp_img) - np.array(out_img))
     num_pix_mismatched = pixelmatch(out_img, comp_img, diff_img, includeAA=True,
                                     alpha=0.33, threshold=0.05)
     fullimg_retval = 0 if np.all(diff_arr == 0) and num_pix_mismatched == 0 else 1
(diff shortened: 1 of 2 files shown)


evrose54:
I would wait on merging this branch into main until @mindyls and @jsolbrig can test this out and see if they like the new functionality. This is definitely more of test to see if you prefer this comparison method opposed to our previous one. If you like it, feel free to make this PR ready to review and get this merged into main.

geoips/compare_outputs.py
```diff
@@ -214,95 +214,42 @@ def images_match(output_product, compare_product, fuzz="5%"):
...
+    from pixelmatch.contrib.PIL import pixelmatch
+
+    LOG.info("**Comparing output_product vs. compare product")
+    out_img = Image.open(output_product)
+    comp_img = Image.open(compare_product)
+    diff_img = Image.new(mode="RGB", size=comp_img.size)
+    diff_arr = np.abs(np.array(comp_img) - np.array(out_img))
+    num_pix_mismatched = pixelmatch(out_img, comp_img, diff_img, includeAA=True,
+                                    alpha=0.33, threshold=0.05)
+    fullimg_retval = 0 if np.all(diff_arr == 0) and num_pix_mismatched == 0 else 1
```
evrose54, line 231: Just as a heads up, I can't figure out one problem when running tests/integration_tests/full_test.sh. I get a 1 on the check_code.sh, which outputs the 'warning' `would reformat /local/home/evrose54/geoips/geoips_packages/geoips/geoips/compare_outputs.py`. It doesn't say what line however. I thought it was because of this line, however I changed it to 
`if if np.all(diff_arr == 0) and num_pix_mismatched == 0: 
fullimg_retval = 0
else:
fullimg_retval = 1
` statement on my local branch, and still got the same output. Does anyone know why this output is occurring? Can't seem to figure this one out. All other tests pass on my end.

evrose54: Found out how to fix it. Resolved.

geoips/compare_outputs.py
```diff
@@ -214,95 +214,46 @@ def images_match(output_product, compare_product, fuzz="5%"):
...
-        "ae",
-        output_product,
-        compare_product,
-        exact_out_diffimg,
-    ]
-
-    LOG.info("**Running %s", " ".join(call_list))
-    fullimg_ret = subprocess.run(
-        call_list, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+    from PIL import Image
```
mindyls, line 220: Per conversation this morning, it is using pillow:

In [2]: import PIL

In [3]: PIL?
Type:        module
String form: <module 'PIL' from 'geoips/lib/python3.10/site-packages/PIL/__init__.py'>
       geoips/lib/python3.10/site-packages/PIL/__init__.py
Docstring:  
Pillow (Fork of the Python Imaging Library)

Pillow is the friendly PIL fork by Jeffrey A. Clark (Alex) and contributors.
    https://github.com/python-pillow/Pillow/

Pillow is forked from PIL 1.1.7.

PIL is the Python Imaging Library by Fredrik Lundh and contributors.
Copyright (c) 1999 by Secret Labs AB.

Use PIL.__version__ for this Pillow version.

;-)

geoips/compare_outputs.py
```diff
@@ -214,95 +214,46 @@ def images_match(output_product, compare_product, fuzz="5%"):
...
+    import numpy as np
+    from pixelmatch.contrib.PIL import pixelmatch
+
+    LOG.info("**Comparing output_product vs. compare product")
+    out_img = Image.open(output_product)
+    comp_img = Image.open(compare_product)
+    diff_img = Image.new(mode="RGB", size=comp_img.size)
+    diff_arr = np.abs(np.array(comp_img) - np.array(out_img))
```
mindyls, line 228: This fails if the images are different sizes - probably just need to catch this.

  File "geoips/geoips/compare_outputs.py", line 228, in images_match
    diff_arr = np.abs(np.array(comp_img) - np.array(out_img))
ValueError: operands could not be broadcast together with shapes (2463,2178,4) (2467,2178,4)

mindyls (approved):
Diff from the below 2 images:

image

image

image

geoips/compare_outputs.py
```diff
@@ -214,95 +214,64 @@ def images_match(output_product, compare_product, fuzz="5%"):
...
+            np.array(out_img).shape,
+            output_product,
+        )
+        LOG.interactive(
+            "    ***   compare_product: %s %s ***",
+            np.array(comp_img).shape,
+            compare_product,
+        )
+        LOG.interactive("    ***************************************")
+        return False
```
mindyls, line 244: This was raising an exception when sizes didn't match - updated to catch the exception, print out the output and comparisons images (and sizes), then return False.

geoips/compare_outputs.py
```diff
@@ -214,95 +214,64 @@ def images_match(output_product, compare_product, fuzz="5%"):
...
+    if fullimg_retval != 0:
         LOG.interactive("    ***************************************")
         LOG.interactive("    *** BAD Images do NOT match exactly ***")
         LOG.interactive("    ***   output_product: %s ***", output_product)
         LOG.interactive("    ***   compare_product: %s ***", compare_product)
-        LOG.interactive("    ***   out_diffimg: %s ***", out_diffimg)
-        LOG.interactive("    ***   exact_out_diffimg: %s ***", exact_out_diffimg)
+        LOG.interactive("    ***   exact dif image: %s ***", exact_out_diffimg)
         LOG.interactive("    ***************************************")
         return False
```
mindyls, line 264: I updated these to be log.interactive.

pyproject.toml
```diff
@@ -81,7 +81,8 @@ dependencies = [
 [project.optional-dependencies]
 test = [
     "pytest",  # Required for unit tests
-    "pytest-cov"  # Reports on test coverage
+    "pytest-cov",  # Reports on test coverage
+    "pixelmatch"  # Reports on test coverage
```
mindyls, line 85: I added the "pixelmatch" requirement to pyproject.toml.

geoips/compare_outputs.py
```diff
@@ -214,95 +214,76 @@ def images_match(output_product, compare_product, fuzz="5%"):
...
+            np.array(out_img).shape,
+            output_product,
+        )
+        LOG.interactive(
+            "    ***   compare_product: %s %s ***",
+            np.array(comp_img).shape,
+            compare_product,
+        )
+        LOG.interactive("    ***************************************")
+        return False
```
mindyls, line 248: I added a "return False" when the sizes don't match - otherwise was raising an exception and not continuing, we just want the non-zero return.

geoips/compare_outputs.py
```diff
@@ -214,95 +214,76 @@ def images_match(output_product, compare_product, fuzz="5%"):
...
+    if fullimg_retval != 0:
         LOG.interactive("    ***************************************")
         LOG.interactive("    *** BAD Images do NOT match exactly ***")
         LOG.interactive("    ***   output_product: %s ***", output_product)
         LOG.interactive("    ***   compare_product: %s ***", compare_product)
-        LOG.interactive("    ***   out_diffimg: %s ***", out_diffimg)
-        LOG.interactive("    ***   exact_out_diffimg: %s ***", exact_out_diffimg)
+        LOG.interactive("    ***   exact dif image: %s ***", exact_out_diffimg)
         LOG.interactive("    ***************************************")
         return False
```
mindyls, line 274: Do we want to add explicit integration tests for both cases (where sizes don't match, and where sizes DO match but images don't match)?  I'm fine either way, maybe it is just a matter of adding an issue to add unit tests for the output comparisons?

mindyls: I added an Issue:
https://github.com/NRLMMD-GEOIPS/geoips/issues/339

geoips/compare_outputs.py
jsolbrig: Yep, this looks good the way that it's working. Such an odd history here...

geoips/compare_outputs.py
jsolbrig: This looks good to me. Thanks for the update!

pyproject.toml
jsolbrig: Does this belong in the test section or should it be moved to the main dependencies?

geoips/compare_outputs.py
jsolbrig: This looks to be handled already.

pyproject.toml
mindyls: It is only needed for tests,  if people want to just run,  they don't need it

jsolbrig (approved):

geoips/compare_outputs.py
```diff
@@ -197,7 +197,7 @@ def get_out_diff_fname(compare_product, output_product, ext=None, flag=None):
 
 
 def images_match(output_product, compare_product, fuzz="5%"):
-    """Use imagemagick compare system command to compare two images.
+    """Use PIL and matplotlib to compare two images.
```
evrose54, line 200: This line should actually be 'Use PIL and Numpy to compare to images'.

evrose54: Also, we can remove the 'fuzz' argument as it is no longer used.

mindyls: I updated the docstring, and made a note that fuzz is currently not implemented.  I think we will want to re-implement that at a later date, to allow slight variations in the imagery still passing (that is the end goal for having the numpy-based comparisons, so we have more control over what constitutes a "match").

@evrose54 Please resolve this conversation if that looks ok to you.

evrose54: That sounds great to me. You're right that it's probably a better idea to leave it in there for the future in case we end up wanting to use it. I'll resolve this now.

PR #331, opened 2023-09-18
Update installation process to use poetry rather than setuptools

SRIKANTH-KUMAR:
# Reviewer Checklist

* [ ] Existing tests pass (integration_tests/full_test.sh, test_all.sh, pytests as appropriate.
      Includes linting checks and documentation build)

  * [ ] No full testing required
---

* [ ] Appropriate unit tests added and pass for new/modified functionality
      
  * [ ] No unit tests required
---

* [ ] Appropriate integration tests added and pass for new/modified functionality
      
  * [ ] No integration tests required.
---

* [ ] Appropriate documentation added for new/modified functionality
      
  * [ ] No documentation required
---

* [ ] Appropriate release notes added for new/modified functionality
      
  * [ ] No release note updates required.
---

# Related Issues
fixes NRLMMD-GEOIPS/geoips#NNN

# Testing Instructions

# Summary

# Output


pyproject.toml
```diff
@@ -1,119 +1,106 @@
...
+    matplotlib = "^3.7.2"           # Base requirement works, version specific to test outputs
+    netcdf4 = "^1.6.4"              # Base requirement
+    numpy = "^1.25.2"               # Base requirement
+    pyresample = "^1.27.1"          # Base requirement Geospatial image resampling 
+    				     	# efficiency improvements >= 1.22.3
+    pyyaml = "^6.0.1"               # Base requirement 
+    scipy = "^1.11.2"  	     # Base requirement, requires openblas 
+    xarray = "^2023.7.0"            # Base requirement
+    
+    astropy = "^5.3.2"              # Used for accounting leap seconds when converting epoch timestamps
```
mindyls, line 52: @jsolbrig I think I would prefer allowing versions to increment, and not tying to the major version number.  I like to see when things break, and then resolve dependency issues as they break, rather than getting way out of date and never noticing.  Thoughts?

pyproject.toml
```diff
@@ -1,119 +1,106 @@
...
+style = "semver"
+#tag-branch = "feature-poetry"
+strict = true
+pattern = "(?x)^((?P<epoch>\\d+)!)?(?P<base>\\d+(\\.\\d+)*)([-._]?((?P<stage>[a-zA-Z]+)[-._]?(?P<revision>\\d+)?))?(\\+(?P<tagged_metadata>.+))?$"
+                         #           also: 3.6.0 incompatible with 0.20.3, but 0.21.0 works
+
+
+[tool.poetry-dynamic-versioning.substitution]
+files = ["*/_version.py"]
+
```
mindyls, line 35: @jsolbrig Do we know if this will work the same way as the current dynamic versioning (setuptools_scm)?

pyproject.toml
```diff
@@ -1,119 +1,106 @@
...
-[tool.setuptools.dynamic]
-entry-points = {file = ["entry-points.ini"]}
-
-[project]
-name = "geoips"
-authors = []
+[tool.poetry]
+# required ones
+name = "geoips"         # geoips package available at the root of the project
+version = "0.0.0"   # this has to be fixed; it's dynamic in setuptools
```
mindyls, line 4: Does this need to be removed now, with the poetry dynamic versioning below?

poetry.lock
```diff
@@ -0,0 +1,3777 @@
+# This file is automatically @generated by Poetry 1.6.1 and should not be changed by hand.
```
mindyls, line 1: @jsolbrig Did we want to include this poetry.lock file?

pyproject.toml
jsolbrig: Actually, poetry uses `"0.0.0"` as a known placeholder so this is correct.

pyproject.toml
jsolbrig: It should. I had @SRIKANTH-KUMAR work pretty hard to make sure that it works the same way. Can you test it to be sure?

When you do `pip install -e .` it will update `_version.py`. 

If you just do `pip install .` it will not updated it because it's not being used from that location. It will be correct in the installed package, though. The same is true if you build a wheel for distribution.

pyproject.toml
jsolbrig: Yeah, I agree that we shouldn't be pinning everything. We should go back and remove all of the pinned versions that aren't already in the current pyproject.toml.

poetry.lock
jsolbrig: Let me look into this a bit more.

jsolbrig: Read this, then we should discuss: https://python-poetry.org/docs/basic-usage/#installing-dependencies

Using poetry.lock feels uncomfortable to me. It also makes it difficult for GeoIPS to be installed into an environment that already contains other packages. I'm not sure if I like it or not.

pyproject.toml
SRIKANTH-KUMAR: The placeholder is needed for the dynamic versioning to work.

pyproject.toml
SRIKANTH-KUMAR: @mindyls @jsolbrig the commented requirement is added in the latest commit on the feature-poetry [branch](https://github.com/SRIKANTH-KUMAR/geoips/commit/ec9650b26382be8e852fd242a5ebdb8fc656eb9b)

poetry.lock
SRIKANTH-KUMAR: Have deleted the poetry.lock file and tried installing using pip; it works as expected. On the sidenote, GeoIPS get installed using poetry install without lock file; it just creates a new one.

pyproject.toml
jsolbrig: This is fixed. Thanks @SRIKANTH-KUMAR

pyproject.toml
jsolbrig: Okay, to further answer your question:

After playing with this some there were some problems that needed to be resolved. After resolving those problems, this works ALMOST identically to setuptools_scm with one exception. This method requires a template file to add versions to. This is going to be a problem because that file must be committed and can accidentally be updated.

See these issues:
- https://github.com/mtkennerly/poetry-dynamic-versioning/issues/147
- https://github.com/mtkennerly/poetry-dynamic-versioning/issues/146

For the time being it might be a good idea to add a CI check or a hook to ensure that `geoips/_version.py` is never updated.

poetry.lock
jsolbrig: I have removed the poetry.lock file. Poetry discusses when it should and should not be used. They suggest that it should be used for top-level packages that won't be used as dependencies for other packages. For "libraries" they suggest not using it.

GeoIPS is a little of both, but since other people will be using GeoIPS as a dependency for their plugin packages, it would be best to not use poetry.lock.

jsolbrig (approved):

jsolbrig:
@mindyls This should be ready for your review.

pyproject.toml
```diff
@@ -1,261 +1,266 @@
...
+readme = "README.md"
+keywords = ["GeoIPS", "Poetry"]
+packages = [{include = "geoips"}]  # not needed as tool.poetry.name is properly set already
+classifiers = [                # // list of PyPI trove classifiers to describe the project
+    "Programming Language :: Python :: 3", ]
+include = [
+    "*.txt",
+    "*.py",
+    "*.yaml",
+    "yaml_configs/*",
```
mindyls, line 23: Does the include now include all levels ? (ie, can we get rid of the "yaml_configs" entries here)?  Or do we need to add back in all of the plugins/*/*/* and schema/*/*/* lines? @jsolbrig

pyproject.toml
```diff
@@ -1,261 +1,266 @@
...
+[tool.poetry-dynamic-versioning]
+enable = true
+vcs = "git"
+metadata = false    # don't include local-version hash; date doesn't appear by deafult
+style = "pep440"    # pep440 is deafult; can rmeove later on; used for : keeping 0.0.8 as the default install until 0.0.9 is out
+#style = "semver"
+#tag-branch = "feature-poetry"
+strict = true
+pattern = "(?x)^((?P<epoch>\\d+)!)?(?P<base>\\d+(\\.\\d+)*)([-._]?((?P<stage>[a-zA-Z]+)[-._]?(?P<revision>\\d+)?))?(\\+(?P<tagged_metadata>.+))?$"
+                         #           also: 3.6.0 incompatible with 0.20.3, but 0.21.0 works
```
mindyls, line 39: That looks like a cartopy comment.

pyproject.toml
```diff
@@ -1,261 +1,266 @@
...
+flake8 = { version = "*", optional = true }
+flake8-rst-docstrings = { version = "*", optional = true }
+flake8-docstrings = { version = "*", optional = true }
+flake8-rst = { version = "*", optional = true }
+prettier = { version = "*", optional = true }
+doc8 = { version = "*", optional = true }
+# Test group
+pytest = { version = "*", optional = true }
+pytest-cov = { version = "*", optional = true }
+pixelmatch = { version = "*", optional = true }
```
mindyls, line 96: Did you catch all the updated versions / packages since this was branched? @jsolbrig I think it was just jsonschema, numexpr, and pixelmatch?

pyproject.toml
```diff
@@ -1,261 +1,266 @@
...
+    "bandit",                # Syntax/security checking
+    "black",                 # Syntax checking and reformatting
+    "pylint",                # Syntax checking
+    "flake8",                # Syntax checking
+    "flake8-rst-docstrings", # numpy docstring checking
+    "flake8-docstrings",     # docstring checking
+    "flake8-rst",            # RST checking
+    "prettier",              # YAML, JSON, etc linting
+    "doc8"                   # RST linting
+]
```
mindyls, line 119: Are all of the dependencies specified twice?  under lint = [, and with the {} above?  @jsolbrig

pyproject.toml
```diff
@@ -1,261 +1,266 @@
...
+    "pytest-cov",    # Reports on test coverage
+    "pixelmatch"
 ]
 debug = [
-    "ipython",
+    "ipython"
 ]
 
-[project.entry-points."geoips.plugin_packages"]
+[tool.poetry.plugins."geoips.plugin_packages"]
```
mindyls, line 129: Does tool.poetry.plugins work the same as project.entry-points?

pyproject.toml
```diff
@@ -1,261 +1,266 @@
...
 wind_radii_transitions = "geoips.plugins.modules.colormappers.winds.wind_radii_transitions"
 
-[project.entry-points."geoips.filename_formatters"]
+[tool.poetry.plugins."geoips.coverage_checkers"]
+center_radius = "geoips.plugins.modules.coverage_checkers.center_radius"
+center_radius_rgba = "geoips.plugins.modules.coverage_checkers.center_radius_rgba"
+masked_arrays = "geoips.plugins.modules.coverage_checkers.masked_arrays"
+numpy_arrays_nan = "geoips.plugins.modules.coverage_checkers.numpy_arrays_nan"
+rgba = "geoips.plugins.modules.coverage_checkers.rgba"
+windbarbs = "geoips.plugins.modules.coverage_checkers.windbarbs"
```
mindyls, line 182: Were there any updates in the plugins in pyproject.toml since this was branched?  @jsolbrig

pyproject.toml
jsolbrig: I'm looking into this now. It looks like this is one spot where Poetry may be nicer than Setuptools. It looks like this will work:
```toml
include = [
    "**/*.py",
    "**/*.yaml",
    "**/*.txt"
]
```

pyproject.toml
jsolbrig: Yes, I was careful to go through and capture everything. It might be good for you to double-check but I'm 99% sure I got everything.

pyproject.toml
jsolbrig: Yes, yes they are... They have to be specified twice. The first time specifies which version to use and that it is optional. The second specifies which "extras" include each package.

pyproject.toml
jsolbrig: Generally, yes, this does work the same as entry points, but notice that the console-script entry points became `[tool.poetry.scripts]`. 

So, there are some difference between Poetry's handling of entry points and Setuptools, but the differences are trivial.

pyproject.toml
jsolbrig: I went through them by hand. I grabbed the version from `main`, sorted each section, then compared against the new pyproject.toml.

Like with the dependencies, it would be great if you could double-check Kumar's and my work.

pyproject.toml
SRIKANTH-KUMAR: This was from matplolib; removed just now on the branch [feature-poetry](https://github.com/SRIKANTH-KUMAR/geoips/commit/bf80d793f664574860ef4e4e4f9c8ba0ca16530e).

pyproject.toml
SRIKANTH-KUMAR: adding from what I did from my end :  numexpr was recently updated (this week) with a constraint <2.8.5 when my build failed. Jeremy asked for this change.

pyproject.toml
SRIKANTH-KUMAR: adding from what I have observed since branched: there was a nomenclature change like coverage_check to coverage_checkers both in namespace and specific plugin. There were others too; all being updated.

pyproject.toml
SRIKANTH-KUMAR: yes

mindyls (approved):
This looks good to me, thanks for responding to my comments, I resolved them all!

Full test is chugging along, everything passing so far.

I did confirm it works with the 1.11.3a0 tag layout (I know I had to spend some time picking a format that worked with setuptools_scm).  They get auto-updated to 1.11.3a0.postN.dev0 (N increments with each commit after the 1.11.3a0 tag):

(geoips) [geoips]$ poetry-dynamic-versioning 
Version: 1.11.3a0.post2.dev0
Files with substitutions:
  - $HOME/geoips/geoips/_version.py

Successfully built geoips
Installing collected packages: geoips
  Attempting uninstall: geoips
    Found existing installation: geoips 1.11.3a0.post1.dev0
    Uninstalling geoips-1.11.3a0.post1.dev0:
      Successfully uninstalled geoips-1.11.3a0.post1.dev0
Successfully installed geoips-1.11.3a0.post2.dev0

pyproject.toml
mindyls: Hmm, is this even a problem, since it doesn't actually work with pip install -e?  I'm going to not worry about this right now, we will throw it in the "don't merge if modified" list so at least we won't accidntally merge to main.

PR #344, opened 2023-09-21
modify update_this_release_note to v1_11_5a0.rst

mindyls:
Identify v1_11_5a0.rst as the next release note/version

Thinking of adding this file so it is clear to everyone which release note should be actively updated.  I just tagged v1.11.3a0, so we are now moving onto v1.11.5a0 for new updates.

The next version is always pre-determined, so use this specific file in the geoips repo to identify the actively developed version for all repos.

Maybe it makes sense for this to be a PR immediately after v1.11.3a0 is tagged?


mindyls:
Was updated with 1.11.3 release

PR #349, opened 2023-09-27
Create default v1_11_5a0.rst in geoips repo

mindyls:
Create default release note in geoips repo

This will be part of the release process going forward, since we are already changing the "update_this_release_note" file, we know geoips repo will always have changes on every version release, so we are safe to auto-generate the new release note itself.

Other repos we do not want to auto-generate the release note, because they will not necessarily have changes / be tagged on every release, but should be safe with geoips repo.

### DIFF
--- a/docs/source/releases/v1_11_5a0.rst
+++ b/docs/source/releases/v1_11_5a0.rst
@@ -1,0 +1,26 @@
+ | # # # Distribution Statement A. Approved for public release. Distribution unlimited.
+ | # # #
+ | # # # Author:
+ | # # # Naval Research Laboratory, Marine Meteorology Division
+ | # # #
+ | # # # This program is free software: you can redistribute it and/or modify it under
+ | # # # the terms of the NRLMMD License included with this program. This program is
+ | # # # distributed WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


docs/source/releases/v1_11_5a0.rst
```diff
@@ -0,0 +1,26 @@
...
+ | # # # Naval Research Laboratory, Marine Meteorology Division
+ | # # #
+ | # # # This program is free software: you can redistribute it and/or modify it under
+ | # # # the terms of the NRLMMD License included with this program. This program is
+ | # # # distributed WITHOUT ANY WARRANTY; without even the implied warranty of
+ | # # # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the included license
+ | # # # for more details. If you did not receive the license, for more information see:
+ | # # # https://github.com/U-S-NRL-Marine-Meteorology-Division/
+
+Version 1.11.5a0 (2023-09-27)
```
mindyls, line 13: Adding conversation to make sure conversation resolution still works the same as well.

cpcamacho (approved):

evrose54 (approved):
This looks good to me!

PR #345, opened 2023-09-22
337 Fixed Scipy import deprecation warning

evrose54:
# Reviewer Checklist

* [x] Required ***existing tests*** pass (ie full_test.sh, others as appropriate)
  * [x] None required (explain if checked, delete line if not checked)
* [x] Required ***unit tests*** added and pass for new/modified functionality
  * [x] None required (explain if checked, delete line if not checked)
* [x] Required ***integration tests*** added and pass for new/modified functionality
  * [x] None required. (explain if checked, delete line if not checked)
* [x] Required ***documentation*** added for new/modified functionality
  * [x] None required (explain if checked, delete line if not checked)
* [x] Required ***release notes*** added for new/modified functionality
  * [x] None required. (explain if checked, delete line if not checked)

https://github.com/NRLMMD-GEOIPS/.github/blob/main/.github/review-template.md

# Related Issues
fixed #337 

# Testing Instructions
Run geoips/tests/integration_tests/full_test.sh to prove that no errors have occurred.

# Summary
This PR has been created to update the 'from scipy.ndimage.interpolation import zoom' to 'from scipy.ndimage import zoom', as the interpolation library was deprecated. No problems exist anymore. This PR also adds the new v1_11_5a0.rst to the release docs.

  - modified: geoips/geoips/plugins/modules/readers/abi_netcdf.py
  - modified: geoips/geoips/plugins/modules/readers/ahi_hsd.py
  - added: geoips/docs/source/releases/v1_11_5a0.rst

PR #343, opened 2023-09-21
322 Replace local path links within documentation with sphinx 'ref' links

evrose54:
# Reviewer Checklist

* [x] Required ***existing tests*** pass (ie full_test.sh, others as appropriate)
  * [x] None required (explain if checked, delete line if not checked)
* [x] Required ***unit tests*** added and pass for new/modified functionality
  * [x] None required (explain if checked, delete line if not checked)
* [x] Required ***integration tests*** added and pass for new/modified functionality
  * [x] None required. (explain if checked, delete line if not checked)
* [x] Required ***documentation*** added for new/modified functionality
  * [x] None required (explain if checked, delete line if not checked)
* [x] Required ***release notes*** added for new/modified functionality
  * [x] None required. (explain if checked, delete line if not checked)

Note: I've yet to add changes into release notes as I'm waiting for the next release_notes.rst file to be created.

https://github.com/NRLMMD-GEOIPS/.github/blob/main/.github/review-template.md

# Related Issues
fixes #321 
fixes #322 

# Testing Instructions
To test the new documentation updates, run this command, replacing $geoips_repopath with your path to the top level
of the geoips folder, and $geoips_pkgname with 'geoips'

$geoips_repopath/docs/build_docs.sh $geoips_repopath $geoips_pkgname html_only

# Summary
This PR had been created to add links to corresponding sections throughout GeoIPS documentation. I may have missed some links, but after checking a few times I think it's all in order. 
Added links to:
  - geoips/docs/source/starter/expert_installation.rst
  - geoips/docs/source/started/extending.rst
  - geoips/docs/source/starter/installation.rst
  - geoips/docs/source/userguide/plugin_extend.rst

# Output
To view the output, download the geoips/build/sphinx/html directory, and copy your index.html file into your browser.

PR #330, opened 2023-09-15
329 Update Sphinx Docs 'User Guide' --> 'Extend GeoIPS With Plugins'

evrose54:
# Reviewer Checklist

* [x] Existing tests pass (integration_tests/full_test.sh, test_all.sh, pytests as appropriate.
      Includes linting checks and documentation build)

  * [x] No full testing required
---

* [x] Appropriate unit tests added and pass for new/modified functionality
      
  * [x] No unit tests required
---

* [x] Appropriate integration tests added and pass for new/modified functionality
      
  * [x] No integration tests required.
---

* [x] Appropriate documentation added for new/modified functionality
      
  * [x] No documentation required
---

* [x] Appropriate release notes added for new/modified functionality
      
  * [x] No release note updates required.
---

https://github.com/NRLMMD-GEOIPS/.github/blob/main/.github/review-template.md

# Related Issues
fixes #79
fixes #329 

# Summary
This PR adds documentation for everything we discussed in the tutorial.
Includes:
  - Products
  - Product Defaults
  - Algorithms
  - Readers
  - Static Sectors
  - Feature Annotators
  - Gridline Annotators
  - Colormappers
  - Output Formatters
  - And Updated Top Level Info for Extending GeoIPS with Plugins

PR #351, opened 2023-09-28
293 Convert compare_outputs.py into an 'output_checkers' interface

### LINKED ISSUE #293: Create interface for "compare_outputs"
# Requested Update

### Description
* Currently, the "compare_outputs.py" module in the main geoips repo contains a number of different functions for comparing different types of outputs.
* This module should be turned into an interface, to allow modularly specifying the various output comparisons.
* This will also allow easily adding more types of output comparisons (ie, pure python based comparisons to remove reliance on imagemagick, as well as non-image based comparisons to reduce reliance on matplotlib/cartopy versions when comparing outputs).
* Currently compare_outputs is not set up as an interface, so there will be some effort required to rearrange it into a format conducive to using as an interface.

### Background and Motivation

### Alternative Solutions
  
### Environment

### Code to demonstrate issue

### Checklist for Completion
* [ ]

evrose54:
# Reviewer Checklist

  * [x] None required -- Fully tested already.
  * [x] None required -- None needed for this branch. Suite of tests cover these updates.
  * [x] None required. -- Current integration tests do the job fine.
  * [x] None required -- Documentation has been added to the code.
  * [x] None required -- Release notes added.

# Related Issues
fixes #293 

# Testing Instructions
To test this, run geoips/tests/integration_tests/full_test.sh. This PR is still a draft and is not ready for merge yet, however the tests are passing at the moment!

# Summary
Currently, GeoIPS compares output products with a python file called 'compare_outputs.py'. While this worked, we really should compare outputs using a GeoIPS interface. This PR has created an 'output_checkers' interface, which removes the need to use the compare outputs python script directly.

Files Modified:

  - geoips/interfaces/__init__.py
  - geoips/interfaces/base.py
  - geoips/plugins/procflows/single_source.py
  - geoips/plugins/procflows/config_based.py
  - pyproject.toml

Files Added:

  - geoips/interfaces/module_based/output_checkers.py
  - geoips/plugins/modules/output_checkers/geotiff_checker.py
  - geoips/plugins/modules/output_checkers/gz_checker.py
  - geoips/plugins/modules/output_checkers/image_checker.py
  - geoips/plugins/modules/output_checkers/netcdf_checker.py
  - geoips/plugins/modules/output_checkers/text_checker.py

Files Deleted:

  - geoips/compare_outputs.py


geoips/plugins/modules/procflows/single_source.py
```diff
@@ -1379,7 +1379,7 @@ def call(fnames, command_line_args=None):
         reader_kwargs = {}
     compare_path = command_line_args["compare_path"]
     output_file_list_fname = command_line_args["output_file_list_fname"]
-    compare_outputs_module = command_line_args["compare_outputs_module"]
+    # compare_outputs_module = command_line_args["compare_outputs_module"]
```
mindyls, line 1382: Ahh, so now it will find installed output checkers for indiviidual data types in any repo.  That is good.

pyproject.toml
```diff
@@ -1,33 +1,32 @@
 [tool.poetry]
 # required ones
-name = "geoips"         # geoips package available at the root of the project
+name = "geoips"                                          # geoips package available at the root of the project
```
mindyls, line 3: What linting are you using for TOML?  We should probably set that up in general, or we're going to end up with a lot of reformatting changes like this...

pyproject.toml
```diff
@@ -259,3 +257,10 @@ flat_sectorfile_parser = "geoips.interface_modules.trackfile_parsers.flat_sector
 static_standard = "geoips.plugins.modules.title_formatters.static_standard"
 tc_copyright = "geoips.plugins.modules.title_formatters.tc_copyright"
 tc_standard = "geoips.plugins.modules.title_formatters.tc_standard"
+
+[tool.poetry.plugins."geoips.output_checkers"]
+geotiff_checker = "geoips.plugins.modules.output_checkers.geotiff_checker"
+gz_checker = "geoips.plugins.modules.output_checkers.gz_checker"
+image_checker = "geoips.plugins.modules.output_checkers.image_checker"
+netcdf_checker = "geoips.plugins.modules.output_checkers.netcdf_checker"
+text_checker = "geoips.plugins.modules.output_checkers.text_checker"
```
mindyls, line 266: This is the only actual change to pyproject.toml, correct?

mindyls (approved):
Merged #345 #351 #343 #330 into a single branch and ran full_test.sh on all of them:

I am good to merge once my conversations/questions are resolved.

geoips/interfaces/module_based/output_checkers.py
```diff
@@ -0,0 +1,73 @@
...
+        )
+
+        output_checkers = [
+            geotiff_checker,
+            image_checker,
+            gz_checker,
+            netcdf_checker,
+            text_checker,
+        ]
+        correct_checker = None
```
mindyls, line 61: Shouldn't these be automatically discovered, without explicitly importing each one?  We want to be able to add new output checkers to other repos, without referencing them here?

geoips/plugins/modules/output_checkers/geotiff_checker.py
```diff
@@ -0,0 +1,113 @@
...
+              * goodcomps, badcomps, compare_strings
+
+        * If None, use geoips.compare_outputs.test_product)
+
+    Returns
+    -------
+    int
+        Binary code: 0 if all comparisons were completed successfully.
+    """
+    retval = co.compare_outputs(compare_path, output_products, test_product_func)
```
mindyls, line 112: Can this just call "geotiffs_match" directly, without going through the general "compare_outputs" module?  I would assume once we are inside a given output checker, we know what the format is/should be, and can just check it directly?

geoips/plugins/modules/output_checkers/utils/compare_outputs.py
```diff
@@ -541,35 +135,42 @@ def test_product(output_product, compare_product, goodcomps, badcomps, compare_s
...
 
-    if is_geoips_netcdf(output_product):
+    if netcdf_checker.correct_type(output_product):
         matched_one = True
         compare_strings += ["GEOIPS NETCDF "]
-        if geoips_netcdf_match(output_product, compare_product):
+        if netcdf_checker.geoips_netcdf_match(output_product, compare_product):
             goodcomps += ["GEOIPS NETCDF {0}".format(output_product)]
         else:
             badcomps += ["GEOIPS NETCDF {0}".format(output_product)]
```
mindyls, line 176: I think this would all be better directly in the interface, so the actual output checkers are standalone.  Ie, from within the interface, atuomatically identify all of the output checkers, then find call "correct_type" for each one until you find the right one.

We should probably eventually have a way to explicitly specify the output checker type, but for now looping through all of them until we find the right one is probably fine (that's what it's doing now)

geoips/plugins/modules/procflows/config_based.py
```diff
@@ -1691,10 +1690,12 @@ def call(fnames, command_line_args=None):
...
         if cpath != "no_comparison":
-            curr_compare_outputs = find_entry_point(
-                "output_comparisons", final_products[cpath]["compare_outputs_module"]
+            from geoips.interfaces.module_based.output_checkers import output_checkers
+
+            output_checker = output_checkers.get_checker(
+                final_products[cpath]["files"][0]
             )
-            curr_retval = curr_compare_outputs(cpath, final_products[cpath]["files"])
+            curr_retval = output_checker.call(cpath, final_products[cpath]["files"])
```
mindyls, line 1698: Maybe this should be "select_checker" or "identify_checker", to differentiate from the standard "get" interface method

Ie, like all the other interfaces, if you do "output_checkers.get("geotiff_checker") it will return the explicitly named geotiff_checker.  We can use that if we want to eventually specify which checker specifically we should use.

We can attach a helper function called 
``checker_plugin_name = output_checkers.identify_checker("<filename>")``
, which will return, say "geotiff_checker", then you can do
``checker_plugin.get(checker_plugin_name)``

geoips/plugins/modules/procflows/single_source.py
```diff
@@ -1856,10 +1856,10 @@ def call(fnames, command_line_args=None):
 
     retval = 0
     if compare_path:
-        from geoips.geoips_utils import find_entry_point
+        from geoips.interfaces.module_based.output_checkers import output_checkers
 
-        compare_outputs = find_entry_point("output_comparisons", compare_outputs_module)
-        retval = compare_outputs(
+        output_checker = output_checkers.get_checker(final_products[0])
+        retval = output_checker.call(
```
mindyls, line 1862: See comment above.  This could either be "output_checkers.find_checker" and return the actual plugin, as you are doing here.  Or separate it out into "output_checkers.identify_checker" and output_checkers.get(<checker_name>)

mindyls (changes requested):
I left a few comments on how I think we might want to restructure the interface slightly.  @jsolbrig  what do you think?

pyproject.toml
evrose54: Yup. Not sure why there are other things included in this diff. This is the only change I made to pyproject.toml

pyproject.toml
evrose54: How can I check that? I'm not sure.

geoips/plugins/modules/procflows/config_based.py
evrose54: That's not a bad idea. Or we can create just one function like "select_checker" as you mentioned above, which automatically gets the checker for us. Seems unnecessary to have it split up into two functions, however I'm happy to implement whichever you think is more appropriate. If we end up wanting to grab a checker via get, the two method approach might be better suited.

geoips/interfaces/module_based/output_checkers.py
evrose54: You're right. That was a brain fart on my side! I'll change that now.

geoips/plugins/modules/procflows/config_based.py
mindyls: Either way is fine. output_checkers.get() is already a built in interface function that is on all interfaces, right @jsolbrig ?

pyproject.toml
mindyls: It is probably within your VSCode settings.  @jsolbrig  ?

geoips/plugins/modules/procflows/single_source.py
evrose54: Updated the function signature to be 'select_checker'

geoips/plugins/modules/output_checkers/geotiff_checker.py
mindyls: I think as long as this ALWAYS imports compare_outputs from the geoips repo that is probably fine (ie, there is no reference to the explicit output_checker plugins within compare_outputs - it uses get_plugins to implicitly find all the output checker plugins in any repo).  We just don't want to go back to how it was when recenter_tc and geoips repos both had completely separate "compare_outputs" Modules...

mindyls: Actually, maybe the "compare_outputs" utility should be attached to the interface itself, rather than a separate python script in output_checkers/utils ?  @jsolbrig should probably chime in on this.

evrose54:
I updated recenter_tc and data_fusion package to use the output_checkers interface

geoips/plugins/modules/output_checkers/geotiff_checker.py
evrose54: I'll wait to update this and the comment below until we hear back from @jsolbrig. Might be a good idea to attach it to the interface, so I'll wait before I change anything in the 'compare_outputs' code.

pyproject.toml
jsolbrig: I think it's probably "Even Better TOML".

geoips/plugins/modules/output_checkers/geotiff_checker.py
jsolbrig: As an update here, we're discussing how best to handle use of methods from inherited BasePlugin classes. Say we define `OutputCheckersBasePlugin.compare_outputs()`, how can that be accessed from within a plugin's `call()` function.

Currently, a plugin's `call()` function is attached to the plugin's class as a static method like this:
```
obj_attrs["__call__"] = staticmethod(getattr(plugin_module, "call"))
```
This makes is so that the resulting plugin object is callable and will execute `plugin_module.call()` when the object is called. So:
```
plg(foo, bar)
```
calls
```
plg.__call__(foo, bar)
```
which calls
```
plugin_module.call(foo, bar)
```

Since this is implemented as a static method, though, it does not provide access to `self` for the object. This means that we can't access any of the plugin object's instance methods. So, if we define `OutputCheckersBasePlugin.compare_outputs()` we can't access it from within `call()`.

There are a few methods to make instance methods available from within `call()`:

1. We could require all plugin modules' `call()` functions to include one additional argument such as `self` or 
    `plugin`, then attach `call()` to the plugin object as an instance method. using:
    ```
    obj_attrs["__call__"] = instancemethod(getattr(plugin_module, "call"))
    ```
    This would allow us to do something like this:
    ```
    def call(self, foo, bar):
        self.compare_outputs()
    ```
2. Rather than requiring `self` or `plugin` as arguments, we could make them optional, then attach the
    `call()` functions to their classes as staticmethods if `self` or `plugin` is absent and as instancemethods if 
    they are present.
3. We could create an additional method of specifying a plugin by specifying it as a class directly rather than 
    converting a plugin module to a class. So, rather than a plugin module, we could specify:
    ```
    class ImageChecker(OutputCheckersBasePlugin):
        def __call__(self, foo, bar):
            self.compare_outputs()
    ```
4. We could also just stop doing module based plugins and require that all python plugins be written as classes 
    in the first place.

Personally, I prefer option 3 or option 4. 
- Option 3 avoids creating more magic and making module-based plugins more confusing. 
- Option 4 gets rid of all of the magic, but makes implementing a new plugin a little bit more daunting for a non-python programmer and requires lots of changes in lots of places.

I think that option 3 is probably the best compromise solution that avoids the most headaches for everyone including people who have already developed plugins for GeoIPS while option 4 would result in GeoIPS being cleaner and easier to work with.

geoips/interfaces/module_based/output_checkers.py
```diff
@@ -0,0 +1,500 @@
...
+            if basename(output_product) in compare_basenames:
+                if test_product_func is None:
+                    test_product_func = self.test_products
+                goodcomps, badcomps, compare_strings = test_product_func(
+                    output_product,
+                    join(compare_path, basename(output_product)),
+                    goodcomps,
+                    badcomps,
+                    compare_strings,
+                )
```
mindyls, line 154: Should "test_product" just be another expected method on each plugin within this class?  So rather than passing in the "test_product_func" function, just call self.test_product ?

geoips/plugins/modules/procflows/config_based.py
```diff
@@ -1691,13 +1690,18 @@ def call(fnames, command_line_args=None):
...
+            for output_product in final_products[cpath]["files"]:
+                output_checker = output_checkers.get_plugin(output_product)
+                curr_retval = output_checker(
+                    output_checker,
+                    cpath,
+                    [output_product],
+                )
+                retval += curr_retval
+                if curr_retval != 0:
+                    failed_compares[cpath] = curr_retval
```
mindyls, line 1704: Ah, so your solution to not having access to the plugin itself within the plugin is to just pass in the plugin explicitly ?  When we decide how to handle this natively within all plugin classes we could presumably remove the "output_checker" argument?

geoips/plugins/modules/output_checkers/text_checker.py
```diff
@@ -0,0 +1,122 @@
...
+
+        * If None, use geoips.compare_outputs.test_product)
+
+    Returns
+    -------
+    int
+        Binary code: 0 if all comparisons were completed successfully.
+    """
+    retval = plugin.compare_outputs(compare_path, output_products, test_product_func)
+    return retval
```
mindyls, line 122: Can we remove the need for "test_product_func" ?  Shouldn't it be assumed now that we have separate output checker plugins that the "test_product_func" is just whatever is defined internally to the current output checker?

Do we just need to specify additional required methods for this interface that determine how to compare the outputs, and how to check if a given output is of the type required to use this output checker?

geoips/interfaces/module_based/output_checkers.py
```diff
@@ -0,0 +1,500 @@
...
+        if not checker_found:
+            raise TypeError("There isn't an output checker built for this data type.")
+        return checker_name
+
+    def get_plugin(self, name):
+        """Get the output checker plugin corresponding to checker_name and return it."""
+        try:
+            return super().get_plugin(name)
+        except PluginError:
+            return super().get_plugin(self.identify_checker(name))
```
mindyls, line 497: So this supports calling get_plugin with either the path to a filename or the name of the plugin?  Do we want to be more explicit here, or is this the "pythonic" way of doing it?

geoips/interfaces/module_based/output_checkers.py
```diff
@@ -0,0 +1,500 @@
...
+            compare_strings += ["GEOIPS NETCDF "]
+            if self.module.outputs_match(self, output_product, compare_product):
+                goodcomps += ["GEOIPS NETCDF {0}".format(output_product)]
+            else:
+                badcomps += ["GEOIPS NETCDF {0}".format(output_product)]
+
+        if not matched_one:
+            raise TypeError(f"MISSING TEST for output product: {output_product}")
+
+        return goodcomps, badcomps, compare_strings
```
mindyls, line 457: I think we should just require "outputs_match" method on all output checker plugins, then we do not have to have the output checker plugin names listed here explictily.

The only unique part of each of these calls to outputs_match is the prefix to the goodcomps/badcomps strings - perhaps we could just use self.name.upper().  If that happens to break anything downstream with parsing the logs or something, we can fix that (I don't think it will have any impact).

mindyls (changes requested):
Just a couple more organizational changes I think and it should be good.  I think it should be fine to require:
* outputs_match
* correct_type
methods on every output checker plugin?  That is what it means to be an output chekcer, you need to have a way to define that a given file is of the current type, and you need to have a way to define how to compare two files of that type?  So should be fine to just require that?  This would allow us to remove all instances of "test_products_func", which would also allow us to remove the explicit list of plugin names from within the output checkers interface module (in the test_products method).  Thoughts?

geoips/interfaces/module_based/output_checkers.py
evrose54: @jsolbrig recommended I implement this function in this manner. However, I'm happy to leave more comments in this function so it's not as "black-magic" as it is currently.

geoips/interfaces/module_based/output_checkers.py
```diff
@@ -0,0 +1,500 @@
...
+
+                * goodcomps, badcomps, compare_strings
+
+            * If None, use geoips.compare_outputs.test_product)
+
+        Returns
+        -------
+        int
+            Binary code: 0 if all comparisons were completed successfully.
+        """
```
mindyls, line 95: @evrose54 Maybe right at the beginning of the "compare_outputs" module just raise an informative error if "outputs_match" or "correct_type" are not defined?  Note this will likely not work for the gzip and gunzip ones (since they are slightly different...), but I *think* it is ok to just handle those explicitly and slightly differently (since they apply to any file format - it's just a way to gunzip before comparing what the type really is, and gzip when you're done).

geoips/interfaces/base.py
```diff
@@ -254,7 +254,7 @@ def plugin_repr(obj):
     return f'{obj.__class__}(name="{obj.name}", module="{obj.module}")'
 
 
-def plugin_module_to_obj(name, module, obj_attrs={}):
+def plugin_module_to_obj(cls, name, module, obj_attrs={}):
```
jsolbrig, line 257: I think we should move this function onto the BaseModuleInterface class in the same way that plugin_yaml_to_obj is attached to BaseYamlInterface. It should also be a classmethod.

geoips/interfaces/base.py
```diff
@@ -662,14 +666,14 @@ def get_plugin(self, name):
                 f" attempting to access the correct plugin name)"
             ) from resp
         # Convert the module into an object
-        return plugin_module_to_obj(name, module)
+        return plugin_module_to_obj(self, name, module)
```
jsolbrig, line 669: If we move plugin_module_to_obj onto this class as a classmethod, this line can be reverted.

geoips/interfaces/base.py
```diff
@@ -662,14 +666,14 @@ def get_plugin(self, name):
...
 
     def get_plugins(self):
         """Get a list of plugins for this interface."""
         plugins = []
         for ep in get_all_entry_points(self.name):
             try:
-                plugins.append(plugin_module_to_obj(ep.name, ep))
+                plugins.append(plugin_module_to_obj(self, ep.name, ep))
```
jsolbrig, line 676: If we move plugin_module_to_obj onto this class as a classmethod, this line can be reverted.

geoips/interfaces/module_based/output_checkers.py
```diff
@@ -0,0 +1,500 @@
...
+
+        return goodcomps, badcomps, compare_strings
+
+
+class OutputCheckersInterface(BaseModuleInterface):
+    """Output Checkers routines to apply when comparing data outputs."""
+
+    name = "output_checkers"
+    required_args = {"standard": {}}
+    required_kwargs = {"standard": {}}
```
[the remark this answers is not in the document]
jsolbrig: Yeah, those should be added to required_args here I think.

geoips/interfaces/module_based/output_checkers.py
```diff
@@ -0,0 +1,557 @@
...
+                    badcomps,
+                    compare_strings,
+                )
+            else:
+                missingcomps += [output_product]
+            final_output_products += [output_product]
+
+            # Make sure we leave things as we found them
+            if rezip is True:
+                self.gzip_product(output_product)
```
mindyls, line 221: Where does rezip get set ?

Perhaps all we want to do here is right at the beginning of this function if the extension is .gz, we gunzip then set rezip=True (so not global), then at the end if rezip is true, we rezip it.  So everything is completely internal to this "compare_outputs" function?

Also, I think what @jsolbrig was suggesting was that we could ALSO gunzip the test output (what is stored in the repo) when we actually do the comparison, then re-zip it - which would prevent the unzipped output sitting in the repo all the time.

geoips/interfaces/module_based/output_checkers.py
```diff
@@ -0,0 +1,557 @@
...
+        from glob import glob
+        from os.path import basename, exists, isdir, isfile, join
+
+        compare_basenames = [basename(yy) for yy in glob(compare_path + "/*")]
+        final_output_products = []
+
+        for output_product in output_products:
+            if isdir(output_product):
+                # Skip 'diff' output directory
+                continue
```
mindyls, line 194: So maybe right here you gunzip each output product.

geoips/interfaces/module_based/output_checkers.py
```diff
@@ -0,0 +1,557 @@
...
+                    "compare path but existing in current output path"
+                )
+            )
+            LOG.info(
+                (
+                    "    W (4-5): MISSINGPROD: Number of products existing in "
+                    "comparepath, but missing in current output path"
+                )
+            )
+
```
mindyls, line 437: I think here at the very end you would have to re-zip all of the files found in "compare_path" as appropriate (so you may need to do some tracking of those in a dictionary or something)

geoips/interfaces/module_based/output_checkers.py
```diff
@@ -0,0 +1,557 @@
...
+        LOG.info(
+            "**************************************************************************"
+            "****"
+        )
+        LOG.info("")
+        from glob import glob
+        from os.path import basename, exists, isdir, isfile, join
+
+        compare_basenames = [basename(yy) for yy in glob(compare_path + "/*")]
+        final_output_products = []
```
mindyls, line 189: Probably right here, if we wanted to implement gzip/gunzip for the test products (the ones stored in the repo), you'd have to loop through them all and gunzip any that end in .gz.  Then you'd probably have to have a dictionary of "rezips" to determine which of the files need to be rezipped at the end.  ie rezip[filename] = True or rezip[filename] = False

mindyls (approved):
Please review and resolve Jeremy's conversations - let me know if any additional issues need to be created resulting from these comments

geoips/plugins/modules/output_checkers/geotiff_checker.py
mindyls: If this just needs an issue opened, let's open the issue, reference it here, then resolve this conversations.

geoips/interfaces/module_based/output_checkers.py
mindyls: I'm ok with this, if that's how we want it to work.

geoips/interfaces/module_based/output_checkers.py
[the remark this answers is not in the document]
mindyls: If we can just add an issue for thhis, reference it here then we can resolve the conversation

geoips/interfaces/base.py
evrose54: I'm going to create an issue for this. I don't want to include too many changes in this PR, and I feel that it would be best to have this separated in another branch. That way, if anything goes wrong, we can track it accordingly.

geoips/interfaces/base.py
evrose54: This issue will also fix your comments below.

geoips/interfaces/base.py
evrose54: See https://github.com/NRLMMD-GEOIPS/geoips/pull/351#discussion_r1355580586

geoips/interfaces/base.py
evrose54: I'll create a separate branch for this. Don't want to include too many changes in this branch, and it will be useful to have this separated just in case something goes wrong and we need to accurately track/source those errors.

mindyls: https://github.com/NRLMMD-GEOIPS/geoips/pull/369

geoips/interfaces/base.py
mindyls: https://github.com/NRLMMD-GEOIPS/geoips/pull/369

geoips/interfaces/base.py
mindyls: https://github.com/NRLMMD-GEOIPS/geoips/pull/369

geoips/interfaces/module_based/output_checkers.py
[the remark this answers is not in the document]
mindyls: https://github.com/NRLMMD-GEOIPS/geoips/issues/370
I created an issue.  Resolving.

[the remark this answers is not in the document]
mindyls: https://github.com/NRLMMD-GEOIPS/geoips/issues/363
Oops, I had already created an issue...

geoips/plugins/modules/output_checkers/geotiff_checker.py
mindyls: This has an issue: https://github.com/NRLMMD-GEOIPS/geoips/issues/362
Resolving.

geoips/plugins/modules/output_checkers/utils/compare_outputs.py
mindyls: This has been addressed.  Resolving.

geoips/plugins/modules/output_checkers/geotiff.py
```diff
@@ -0,0 +1,104 @@
...
+    bool
+        True if it is a geotiff file, False otherwise.
+    """
+    if splitext(fname)[-1] in [".tif"]:
+        return True
+    return False
+
+
+def outputs_match(plugin, output_product, compare_product):
+    """Use diff system command to compare currently produced image to correct image.
```
mindyls, line 45: One more stupid question, then I am good.
@jsolbrig @evrose54 are we good with the "required_methods" names of "correct_type" and "outputs_match" ?

Fine with me, just want to make sure we think about the names before we formalize/finalize them.

mindyls (approved):
One more comment to resolve, then I am good to go!

geoips/plugins/modules/output_checkers/geotiff.py
evrose54: The only change I would _maybe_ recommend is changing _correct_type_ to _correct_product_type_ or _correct_output_type_. Up to you though. I'm fine with either!

mindyls: Or "correct_file_type" since it's not like a python type or anything... I'll leave this one up to @jsolbrig , he is particular about names :-)

mindyls: @evrose54 per @jsolbrig , Update to "correct_file_format", and open an issue to investigate "filetype" module to use rather than exclusively checking file extensions for determining file types.

THEN YOU CAN RESOLVE AND WE CAN MERGE!

cpcamacho (approved):

Issue #357, opened 2023-10-04
Create module to convert "geoips" dictionaries of xarray datasets into DataTree


mindyls:
# Requested Update

### Description

Add ``geoips/xarray_utils/datatree.py`` module to start with that contains a function that takes in a "geoips" dictionary of xarray datasets, and returns an xarray datatree object.

### Background and Motivation
* Supports #318 

### Additional Details
from geoips.interfaces import readers
amsr2_reader = readers.get_plugin('amsr2_netcdf')
amsr2_reader('<path_to_amsr2_test_data>')

### Alternative Solutions
* Create a reader that returns DataTree (similar to how the satpy reader will work), though this may use the xarray_utils module anyway.

### Checklist for Completion
* [ ] Function successfully takes in output of geoips reader, and returns an xarray DataTree

PR #359, opened 2023-10-04
358 datatree and pytest 

AJThorpe:
# Reviewer Checklist

  * [ ] None required (explain if checked, delete line if not checked)
* [ ] Required ***unit tests*** added and pass for new/modified functionality
  * [ ] None required (explain if checked, delete line if not checked)
  * [ ] None required. (explain if checked, delete line if not checked)

# Related Issues
fixes NRLMMD-GEOIPS/geoips#357

# Testing Instructions
* Download and run pytest on geoips/xarray_utils/test_dtree.py

# Summary

* Test conversion of dictionary of xarrays to datatree by using datatree
* Build function to convert and check datatree conversion
* Built initial unit test version of datatree conversion script

# Output
* When run locally, pytest produces:

```
code/test/test_dtree.py::test_xarray_to_datatree PASSED                                                                      [100%]

========================================================= warnings summary =========================================================
code/test/test_dtree.py::test_xarray_to_datatree
  <frozen importlib._bootstrap>:241: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 96 from PyObject

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================== 1 passed, 1 warning in 27.19s ==================================================
```


.github/workflows/check-code.yaml
```diff
@@ -127,3 +127,32 @@ jobs:
             echo "::error::due to bandit violations ${ret##*:}"
             exit 1
           fi
+  check_code_pytest:
```
mindyls, line 130: @AJThorpe I think we should create a separate workflow for this, maybe unit-tests.yaml?

docs/source/releases/v1_10_0.rst
```diff
@@ -95,6 +95,13 @@ Was not initially created
   new: schema/product_defaults/interpolator.yaml
   new: schema/product_defaults/specs/interpolator.yaml
 
+Added dictionary xarray to datatree conversion
```
mindyls, line 98: This can go in v1_11_7a0.rst now

geoips/xarray_utils/xr_to_dtree.py
```diff
@@ -0,0 +1,27 @@
...
+
+import logging
+from datatree import DataTree
+
+LOG = logging.getLogger(__name__)
+
+
+def xarray_to_datatree(xarray_dict):
+    """Convert a flat (non nested) dictionary of xarrays to DataTree format"""
+    xarray_datatree = DataTree.from_dict(xarray_dict)
```
mindyls, line 24: wow, this is pretty straightforward, huh...

tests/test_pytest/test_dtree.py
```diff
@@ -0,0 +1,43 @@
...
+    xarray_dict = amsr2_reader(fnames)
+    return xarray_dict
+
+
+def test_xarray_to_datatree(load_testfiles):
+    """Convert a xarray dictionary to a datatree"""
+
+    #check that the datatree is equal to the dict-xarray
+    xarray_datatree = DataTree.from_dict(load_testfiles)
+    assert xarray_datatree == load_testfiles
```
mindyls, line 41: should we call the xr_to_dtree function here?  And do we want to ensure the structure of the resulting dtree is "correct"?

We also might want to think about using a different file than AMSR2, AMSR2 is probably not the fastest one to read (I expect we want these unit tests to run fairly quickly)

mindyls (commented):
I left a couple comments, hopefully you can see them.

geoips/xarray_utils/xr_to_dtree.py
AJThorpe: Yes, I could add a check for a nested dictionary of xarrays if needed...?

.github/workflows/check-code.yaml
jsolbrig: If we do move this to a different file, maybe this file should be renamed to `lint.yaml`. I think we might end up with:
- `run_linters.yaml`
- `run_unit_tests.yaml`
- `run_integration_tests.yaml`

Having separation will let us more easily control when we want different tests to run.

What do you think of the file naming @mindyls?

geoips/xarray_utils/xr_to_dtree.py
jsolbrig: @mindyls Do we ever have tested dictionaries of xarrays? I hope not!

tests/test_pytest/test_dtree.py
jsolbrig: Yes, unit tests are intended to test the actual GeoIPS code so this should be testing `xr_to_dtree`.

How long does this test take to run?

Generally, though, I think it would be better to avoid using any actual test data in the unit tests. They should be able to run without actual test data. Is there a good way for us to set up a dummy xarray dictionary directly in this script to test against?

.github/workflows/check-code.yaml
mindyls: I think
run-linters.yaml
run-unit-tests.yaml
run-integration-tests.yaml
I just renamed to make them all have '-' in the name!

geoips/xarray_utils/xr_to_dtree.py
mindyls: I don't think we do.  @AJThorpe I think we might need to do something to transfer the metadata that is currently stored in the "METADATA" dictionary key, because I expect that could be stored at the top level of the datatree directly, rather than in a  full separate xarray Dataset?

Do you want to see what that might look like?

tests/test_pytest/test_dtree.py
mindyls: @AJThorpe can we create a dummy geoips-like dictionary of xarrays directly in this unit test, then pass it to xr_to_dtree, then make sure the output is what we expect?

mindyls: Oh,.  You already did it. :-)

.github/workflows/check-code.yaml
```diff
@@ -133,3 +133,32 @@ jobs:
...
+        with:
+          repository: NRLMMD-GEOIPS/geoips
+          ref: main
+          token: ${{ secrets.GEOIPS_TOKEN }}
+          path: geoips_dev_utils
+          sparse-checkout: |
+            tests/utils
+      - name: Run code check script pytest
+        run: |
+          pytest -v geoips/xarray_utils/test_dtree.py
```
mindyls, line 163: I think this should just run pytest tests/test_pytest
Right @jsolbrig ?

Then it would run all the pytests in any repository?

.github/workflows/check-code.yaml
```diff
@@ -133,3 +133,32 @@ jobs:
...
+        uses: actions/checkout@v3
+      - name: Checkout geoips main branch
+        uses: actions/checkout@v3
+        with:
+          repository: NRLMMD-GEOIPS/geoips
+          ref: main
+          token: ${{ secrets.GEOIPS_TOKEN }}
+          path: geoips_dev_utils
+          sparse-checkout: |
+            tests/utils
```
mindyls, line 160: I believe this is not needed for pytests, since it is just running pytest in the tests/test_pytest directory, and not using the geoips testing utilities.

geoips/xarray_utils/xr_to_dtree.py
AJThorpe: Yes, understood, I'll add that as the root group in the code

.github/workflows/check-code.yaml
mindyls: @AJThorpe do you want to rename check-code.yaml to run-linters.yaml, then create a new run-unit-tests.yaml file, and move this pytest block to that file?  We do not have any integration tests yet.

mindyls: @AJThorpe @jsolbrig I think we should remove this block of code from the workflow, so only the xr_to_dtree.py, test_dtree.py, and release note files are updated.  We'll address the workflows in a separate PR.

Once this file is reverted, we should be able to just merge this PR with the basics of datatrees implemented.

geoips/xarray_utils/xr_to_dtree.py
mindyls: @AJThorpe Did you make this update?

tests/test_pytest/test_dtree.py
```diff
@@ -0,0 +1,53 @@
...
+    """Generate test xarrays for testing."""
+    # only want a small sample for testing
+    xarray_dict = {
+        "test_{}".format(k): xarray.Dataset(
+            data_vars=dict(temp=(["x", "y"], np.empty((2, 2)))),
+            coords=dict(lon=(["x", "y"], np.empty((2, 2)))),
+        )
+        for k in range(0, 4)
+    }
+    return xarray_dict
```
mindyls, line 36: @jsolbrig Is that the sort of thing you are looking for with pytests?

.github/workflows/check-code.yaml
jsolbrig: Sounds good to me.

jsolbrig (approved):

.github/workflows/check-code.yaml
mindyls: Address this when implemented (removing from workflow for now)

.github/workflows/check-code.yaml
mindyls: Will address this when implemented - removing from workflow in this PR.

mindyls (approved):
Looks good to me, as long as the pytest looks good to @jsolbrig

tests/test_pytest/test_dtree.py
jsolbrig: Yes I think this looks good. The fixture may not be needed at the moment since it is only getting used once, but I don't see that as a problem. It makes it easier to add other tests later if we want.

jsolbrig (approved):

PR #365, opened 2023-10-10
365 add flake8 and black config files

### LINKED ISSUE #364: Point to flake8 and black config files in check code script
# Requested Update

### Description
* Create config files for flake8, black, bandit in either geoips repo or .github repo
* Point to those config files from the tests/utils/check_code.sh script
* Point to those config files from .vscode repository ?
* Point to those config files from documentation

### Background and Motivation

### Alternative Solutions
  
### Environment

### Code to demonstrate issue

### Checklist for Completion
* [ ]

AJThorpe:
# Reviewer Checklist

* [x] Running ```check_code.sh``` on personal computer for the black and flake8 tests with configuration files in ```.config/``` produces the same output with the normal CLI 

# Related Issues
fixes #364 

# Testing Instructions
* [x] Test running ```check_code.sh``` on personal computer for the black and flake8 tests with configuration files in ```.config/``` produces the same output with the normal CLI 
# Summary

# Output


mindyls (approved):
This looks good to me!

tests/utils/check_code.sh
```diff
@@ -79,7 +79,7 @@ if [[ "$test" == "black" || "$test" == "all" ]]; then
     #       substrings directly)
     echo "CALLING TEST:"
     echo "black --check --extend-exclude _version.py --extend-exclude /lib/ --extend-exclude _docs/ --extend-exclude geoips_dev_utils/ $path"
-    black --check --extend-exclude _version.py --extend-exclude /lib/ --extend-exclude _docs/ --extend-exclude geoips_dev_utils/ $path
+    black --config $path/.config/black $path
```
mindyls, line 82: Hmm, this will not work from other repos.  How do we want to handle referencing the geoips repo linter config files from other repos, without having to explicitly reference $GEOIPS_PACKAGES_DIR, or having to check identical config files into every repo?

jsolbrig: Unfortunately, I think the best option is to check them into every repo. One option might be to rely on something like this: https://github.com/marketplace/actions/repo-file-sync-action.

I really wish there was a way to maintain this kind of thing more easily, but I don't really think there is...

AJThorpe:
@jsolbrig Sounds good, I'll reformat it

tests/utils/check_code.sh
```diff
@@ -57,6 +57,13 @@ else
 fi
 
 path=$2
+if [[ -d "./geoips_dev_utils/.config" ]]; then
+    # This is for GitHub Actions, geoips repo cloned as "geoips_dev_utils"
+    CONFIG_PATH=./geoips_dev_utils/.config
+else
+    # This is for local development clone
+    CONFIG_PATH=$GEOIPS_PACKAGES_DIR/geoips/.config
+fi
```
mindyls, line 66: @AJThorpe Perhaps we leave them in the .config directory, update this to reference either $GEOIPS_PACKAGES_DIR/geoips (local development checkout) or geoips_dev_utils/.config (github actions checkout), and call it a day?  So, it will not be automatically referenced from flake8 without passing the path to the config file - but that may be ok (basically we are giving the people the tools to allow them to call flake8/black with our settings, but not forcing the repo to do it by default ?)

tests/utils/check_code.sh
mindyls: @jsolbrig I think we do not treat these as official repo-wide configs, we treat them as reference configs that develiopers can use to configure their linters.  So, we will explictily point to $GEOIPS/.config/black, etc when we run black, etc from our "check_code.sh" script, we can point to that from the .vscode settings iin our vscode repo (since we can assuming geoips repo will be cloned if someone is actively developing), and anyone else can point to them from wherever they want to ensure consistent linting.

.config/.flake8
```diff
@@ -0,0 +1,10 @@
+[flake8]
+max-line-length=88
+count=True
+ignore=E203,W503,E712
+extend-exclude=_version.py,lib,*_docs,geoips_dev_utils
+docstring-convention=numpy
+rst-roles=class,func,ref
+rst-directives=envvar,exception
+rst-substitutions=version
```
mindyls, line 9: Perhaps rename this back to .config/flake8 - since it doesn't really need to be hidden if it is not the "official" flake8 config (though I don't think it matters)

tests/utils/check_code.sh
mindyls: @jsolbrig Thoughts on how else we could handle this for github actions vs local clone ?

.config/.flake8
jsolbrig: I agree with Mindy here.

tests/utils/check_code.sh
jsolbrig: Could we just find the config directory this way?

```bash
SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
GEOIPS_DIR=$(dirname "$(dirname $SCRIPT_DIR)")
CONFIG_DIR="${GEOIPS_DIR}/.config"
```

mindyls: Oh yeah,  that should work,  it will always be relative to the check code script that is currently running...

jsolbrig (approved):

tests/utils/check_code.sh
mindyls: After discussion - these are just resources.  Let people grab these files and use them to assist in their development if they want.  The "official" test is runnig check_code.sh, which references the latest versions of these.

tests/utils/check_code.sh
mindyls: Used `dirname $0`/../../.config

PR #375, opened 2023-10-16
370 reader unit tests

AJThorpe:
# Reviewer Checklist

PR author: Please ensure you meet all of the below requirements, and check boxes appropriately.

Reviewers: Please confirm all required testing/documentation has been completed prior to approving.

* [x] NO REQUIRED ***existing tests*** (No modified full test code)
* [x] Required ***unit tests*** added and pass for new/modified functionality
* [ ] NO REQUIRED ***unit tests*** (explain why not required)
* [x] Required ***release notes*** added for new/modified functionality

https://github.com/NRLMMD-GEOIPS/.github/blob/main/.github/review-template.md

# Summary
Created: [tests/test_pytest/reader_test.py](https://github.com/NRLMMD-GEOIPS/geoips/compare/370-unit-tests?expand=1#diff-91c3ec97bf4e44448b91b8d5b6a6b472a4333df603221248cd737168e81c88b6)
- Tests different readers with the full_install data, xfail for readers with no data given
- Might be useful to look into what cli we want for pytest to produce a useful output, at the moment for summaries I use ``` pytest -q -s --tb=no ```
- Please check if any additional readers could be added or if input files could be added

# Output
Summary:
```
7 passed, 22 xfailed, 7 warnings in 101.68s (0:01:41)
```

PR #376, opened 2023-10-16
Create Image Output Checker Unit Test

### LINKED ISSUE #374: Create Unit Test Suite for Output Checkers Interface
# Requested Update

### Description

With the new addition of the output_checkers interface, there is a need to add unit testing, to make sure everything is working as expected. This comes from https://github.com/NRLMMD-GEOIPS/geoips/pull/367#discussion_r1358900636, which documents the conversation that brought up this issue. 

### Background and Motivation

Output_checkers are tested in the tests/integration_tests/full_test.sh, however these tests are limited, and we would actually like to test the functionality of output_checkers, rather than just their output.

### Alternative Solutions

Leave the code as is.

evrose54:
# Reviewer Checklist

# Related Issues
fixes #374 

# Testing Instructions
Run `pytest tests/test_pytest/output_checkers/test_output_checkers.py` to test. This will say tests pass upon success.

# Summary

Currently, GeoIPS has no unit tests for the output_checkers interface and largely any
GeoIPS interface available. This update adds pytest-based unit tests for GeoIPS image
output_checkers. There will be more unit tests for the remaining output checkers in a
future update. Note, the other files changes have not actually been modified. They've 
just been placed in their appropriate location regarding the new unit test directory structure.

    added: geoips/test/unit_tests/plugins/module/output_checkers/test_output_checkers.py

# Output
Here are some example outputs after running the test script.

diff_test_output_exact_output_img_lenient0

diff_test_output_exact_output_img_lenient1

diff_test_output_exact_output_img_lenient2


tests/test_pytest/test_output_checkers.py
```diff
@@ -0,0 +1,53 @@
...
+
+from geoips.interfaces import output_checkers
+from geoips.commandline import log_setup
+
+
+log_setup.setup_logging()
+image = output_checkers.get_plugin("image")
+savedir = str(__file__).replace(
+    "geoips/tests/test_pytest/test_output_checkers.py", "test_data/test_images/pytest/")
+thresholds = ["lenient", "medium", "strict"]
```
mindyls, line 26: Didn't we change this to floats?  Just pass in the "threshold" argument value to pixelmatch?

tests/test_pytest/test_output_checkers.py
```diff
@@ -0,0 +1,53 @@
...
+from PIL import Image
+import numpy as np
+
+from geoips.interfaces import output_checkers
+from geoips.commandline import log_setup
+
+
+log_setup.setup_logging()
+image = output_checkers.get_plugin("image")
+savedir = str(__file__).replace(
```
mindyls, line 24: Do we want to in general use $GEOIPS_OUTDIRS/scratch for these sorts of things?

tests/test_pytest/test_output_checkers.py
```diff
@@ -0,0 +1,53 @@
...
+    "geoips/tests/test_pytest/test_output_checkers.py", "test_data/test_images/pytest/")
+thresholds = ["lenient", "medium", "strict"]
+
+
+def yeild_images():
```
mindyls, line 30: yield

tests/test_pytest/test_output_checkers.py
```diff
@@ -0,0 +1,53 @@
...
+            output_img = Image.fromarray((output_arr * 255).astype(np.uint8))
+            comp_path = savedir + "comp_img_" + threshold + str(i) + ".png"
+            output_path = savedir + "output_img_" + threshold + str(i) + ".png"
+            comp_img.save(comp_path)
+            output_img.save(output_path)
+            yield (comp_path, output_path)
+
+
+@pytest.mark.parametrize("compare_path, output_path", yeild_images())
+def test_image_comparisons(compare_path, output_path):
```
mindyls, line 51: Do we want to set something up here like @AJThorpe did for the readers, where all the output checkers are included, but the ones that have not been implemented yet wil raise pytest.xfail?

mindyls (commented):
Left some comments

docs/source/releases/v1_11_7a0.rst
```diff
@@ -141,3 +142,19 @@ require changes to the output products as well.
     modified: tests/yaml_configs/abi_test.yaml
     modified: tests/yaml_configs/abi_test_low_memory.yaml
 
+Testing Updates
```
mindyls, line 145: I would say perhaps we *do* update recenter_tc to add a pytest for that output chekcer - so we can see how we want to tie the pytests together for various interfaces (if at all).

tests/test_pytest/test_output_checkers.py
```diff
@@ -0,0 +1,54 @@
...
+# # # Naval Research Laboratory, Marine Meteorology Division
+# # #
+# # # This program is free software: you can redistribute it and/or modify it under
+# # # the terms of the NRLMMD License included with this program. This program is
+# # # distributed WITHOUT ANY WARRANTY; without even the implied warranty of
+# # # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the included license
+# # # for more details. If you did not receive the license, for more information see:
+# # # https://github.com/U-S-NRL-Marine-Meteorology-Division/
+
+"""Test all Output Checker plugins."""
```
mindyls, line 13: Also, @AJThorpe @evrose54 @jsolbrig , do you think we want to organize the test_pytest directory differently at all, as we are getting things set up to add more and more pytests?

test_pytest/test_all_plugins
* from any interface ?
test_pytest/test_plugin_schema
* This is specifically testing the schema, not the plugins themselves
test_pytest/test_utilities
* things like dtree, etc ?

tests/test_pytest/test_output_checkers.py
evrose54: This has yet to be merged in.

tests/test_pytest/test_output_checkers.py
evrose54: I was thinking that, however, I'm not sure if you can access environmental variables like that directly within python. 
Ie, I tested in IPython `print($GEOIPS_PACKAGES_DIR)`, which raised an error that it didn't exist.

tests/test_pytest/test_output_checkers.py
evrose54: Ooops

tests/test_pytest/test_output_checkers.py
evrose54: I can set it up in that fashion if you prefer it that way! Probably a good idea.

docs/source/releases/v1_11_7a0.rst
evrose54: Ok, I can look into that in the future. However, this will essentially be tested the same as the text image checker, so I'm not actually sure if it's necessary, considering it uses the text_checker `outputs_match` function.

tests/test_pytest/test_output_checkers.py
evrose54: In my opinion, I'd split up the directory by interfaces. Ie:
  - test_pytest/
      - readers/
      - output_checkers/
      - ...

evrose54: That way, if we had individual tests for plugins that we didn't want to combine in one file, we could have an organized directory pointing to each correlated test.

tests/test_pytest/test_output_checkers.py
jsolbrig: Generally, I think that should be fine. The data being written there, though, would it get cleaned up or would it stick around? Or would it get overwritten the next time that is called? I’m fine with that directory containing some stuff, but I’m concerned about it growing over time.

evrose54: It's overwritten each time. Max of 27 images in that directory at all times.

tests/test_pytest/test_output_checkers.py
evrose54: Fixed that.

tests/test_pytest/test_output_checkers.py
jsolbrig: Okay, yeah, that sounds fine to me. 

Thinking about this, do you see any reason to retain the files for tests that pass? Maybe it would be a good idea to clean up the scratch files for any tests that pass?

tests/test_pytest/test_output_checkers.py
evrose54: This was merged an I've since updated this.

tests/test_pytest/test_output_checkers.py
evrose54: I do see a reason. This way, we can visually check that each file passes or fails in the way we expect. For other output checkers I would be fine removing these files, however for images specifically, I see the need to keep them.

tests/test_pytest/test_output_checkers.py
evrose54: I added that functionality for the future. If you're fine with its structure, feel free to resolve.

tests/test_pytest/test_output_checkers.py
jsolbrig: Okay, keeping them is fine by me so long as they're being overwritten. I just want to be sure that directory won't grow over time.

evrose54: Unless we end up changing the filenames dynamically, it will never grow. Going to resolve this for now.

tests/test_pytest/test_output_checkers.py
jsolbrig: Yes, this sounds like a good structure to me. Directories for each interface.

mindyls: Do we want to have another level
test_pytest
* test_plugins
  * readers
  * output_checkers
* test_plugin_schema
  * ...
* test_utilities
  * xarray_utils
  * mpl_utils
  * ???

docs/source/releases/v1_11_7a0.rst
mindyls: I still think we should add the test to recenter_tc, because it may inform how we set up the base test.  @jsolbrig what do you think?

tests/test_pytest/test_output_checkers.py
```diff
@@ -0,0 +1,64 @@
...
+import numpy as np
+from os import environ
+
+from geoips.interfaces import output_checkers
+from geoips.commandline import log_setup
+
+
+log_setup.setup_logging()
+image = output_checkers.get_plugin("image")
+savedir = str(environ["GEOIPS_PACKAGES_DIR"]) + "/test_data/test_images/pytest/"
```
mindyls, line 25: Do we want this to be environ["GEOIPS_OUTDIRS"]/scratch ?  I don't think we want pytests writing in $GEOIPS_PACKAGES_DIR. @jsolbrig

tests/test_pytest/test_output_checkers.py
mindyls: https://github.com/NRLMMD-GEOIPS/geoips/pull/375/files#r1361170646
Something like that, where maybe all the output checkers are included in the same dictionary, and for any that have [] specified, raise xfail before trying to run them.  So you might need a higher level test here (not specifically test_image_comparison, but test_output_checkers or something, where it will loop through all the output checkers).

evrose54: I'll work on implementing this.

evrose54: Check this now to see if you like it.

tests/test_pytest/test_output_checkers.py
jsolbrig: Yeah, I think this is right.

tests/test_pytest/test_output_checkers.py
evrose54: I'll leave this up to Jeremy. Generally, I prefer a shorter directory structure, meaning I'd probably have schemas, modules, and yamls as top-level directories, then sort them appropriately by their plugin family/interface. Ie:
  - test_pytest/
      - modules/
          - algorithms
          - ...
      - yamls/
          - products
          - ...
      - schemas/
          - ...

docs/source/releases/v1_11_7a0.rst
mindyls: recenter_tc test added!

mindyls (approved):
This looks good to me.  Discussion about implementing the test methods directly on the plugins (and just having the framework for calling all the tests in the tests directory), but that will be handled in a separate PR, but we will merge this initially.
https://github.com/NRLMMD-GEOIPS/geoips/issues/378

tests/unit_tests/plugins/module/output_checkers/test_output_checkers.py
```diff
@@ -0,0 +1,130 @@
...
+from geoips.commandline import log_setup
+
+
+log_setup.setup_logging()
+
+
+class TestOutputCheckers:
+    """TestOutputChecker class, defining methods as well."""
+
+    savedir = str(environ["GEOIPS_PACKAGES_DIR"]) + "/test_data/test_images/pytest/"
```
jsolbrig, line 29: Probably should use `__file__` to get the location for this rather than using an environment variable.

mindyls:
This is subsumed by this PR, which moves plugin-specific values onto the plugins themselves:
https://github.com/NRLMMD-GEOIPS/geoips/pull/379

PR #380, opened 2023-10-22
Remove circular imports and handle products correctly in create_plugin_registry

mindyls:
Updated create_plugin_registry to use the interface methods in order to properly account for the compound product names for product plugins (ie, source_name, product_name).

Note this caused circular imports if attempting to auto-run create_plugin_registry directly from geoips on runtime (since we are now using geoips functions within create_plugin_registry).  I got around this by no longer attempting to auto-run - just raise a PluginRegistryError if no registered plugins are found.  We could probably figure out how to auto-run it again at runtime if we wanted.

### DIFF
--- a/geoips/geoips_utils.py
+++ b/geoips/geoips_utils.py
@@ -17,11 +17,9 @@ from copy import deepcopy
 import sys
 import yaml
 import logging
-from glob import glob
 from importlib import metadata, resources
 
-from geoips.errors import EntryPointError
-from geoips.filenames.base_paths import PATHS as gpaths
+from geoips.errors import EntryPointError, PluginRegistryError
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


geoips/interfaces/base.py
```diff
@@ -330,10 +330,18 @@ def __new__(cls):
...
     def __init__(self):
         """YAML plugin interface init method."""
-        # self._unvalidated_plugins = self._create_unvalidated_plugins_cache(
-        #     load_all_yaml_plugins()
-        # )
-        self._unvalidated_plugins = load_all_yaml_plugins()
+        try:
+            self._unvalidated_plugins = load_all_yaml_plugins()
+        except PluginRegistryError:
+            self._unvalidated_plugins = {}
```
mindyls, line 336: NOTE: This will just continue on if load_all_yaml_plugins raises an error.  This was to prevent create_plugin_registry from failing when attempting to import geoips.interfaces.products, before even getting to the point of writing out the registered_plugins.yaml files.

geoips/geoips_utils.py
```diff
@@ -78,78 +76,31 @@ def load_all_yaml_plugins():
 
     reg_plug_path = resources.files("geoips") / "registered_plugins.yaml"
     if not os.path.exists(reg_plug_path):
-        from . import create_plugin_registry
-
-        create_plugin_registry.main()
+        raise PluginRegistryError(
+            f"Plugin registry {reg_plug_path} did not exist, "
+            "please run 'create_plugin_registry'"
+        )
```
mindyls, line 82: Note this is where we would end up with circular imports if we tried to run create_plugin_registry during runtime (since we now need geoips.interfaces.products utilities in order to find the correct products to populate the registered_plugins.yaml files).

geoips/interfaces/base.py
jsolbrig: This looks fine for now. To resolve this, we should create a PluginRegistry class that performs a delayed load for plugins. That would allow us to avoid calling load_all_yaml_plugins when we run create_plugin_registry.

PR #392, opened 2023-10-30
Create a PluginRegistry Class

### LINKED ISSUE #381: Create Unit Tests for New Plugin Registry Update
# Requested Update

### Description
With the addition of create_plugin_registry (see #240), we should add unit tests to make sure everything is processed as we expect. This was a large update touching every package that interacts with geoips, and geoips itself, so it would be a great idea things are working as expected.

### Background and Motivation
The new Unit Tests PRs (#379 ...) should provide a good backbone of how to create Unit Tests for this PR. Eventually, we want to have unit tests for all interfaces, and the functionality that interacts with them.

### Alternative Solutions
Leave the code as is and always run tests/integration_tests/full_test.sh. Slow.

### LINKED ISSUE #389: Make better use of the plugin registry
# Requested Update

### Description
Now that we have a plugin registry, we should make better use of it. Currently we call `load_all_yaml_plugins` every time a YAML-based interface is instantiated. We also don't make use of the registry from the module-based interfaces.

The main philosophy behind this issue is:
- If we can use the registry rather than accessing plugins through another means, use the registry.
- If we only need information from the registry and don't need the actual plugin, use the registry.
- If we need to load a plugin, load only that plugin and use the new PluginRegistry class (discussed below) to do the loading.

We should create a new `PluginRegistry` class that takes the place of the `load_all_yaml_plugins` function and the entry-point access to plugins used in the module-based interfaces. This class should:
- provide a property that loads all plugin information from registries and supplies it to the caller, but does not load that information until actually requested.
- (maybe) provide a function that returns a single plugin's registry information.
  - Alternatively, maybe just index into the registry property `plugin_registry.plugins[interface][plugin_name]`.
- provide a function that returns a list of all of the plugin information for a particular interface.
  - This would be used from `Interface.list_plugins()` and will be used by the, eventual, CLI for listing available plugins and their information.

This should be a pretty simple class that looks like this:

```
class PluginRegistry:
    def __init__(self, _test_registry_files=[]):
        # Use this for unit testing
        if _test_registry_files:
            self.registry_files = _test_registry_files
        # Use this for normal operation and collect the registry files
        else:
            self.registry_files = # Collect the paths to the registry files here
        ...
    @property
    def registered_plugins(self):
        # Load the registries here and return them as a dictionary
        if not hasattr(self, "_registered_plugins"):
            self._registered_plugins = # logic to load registries goes here
        return self._registered_plugins
    def get_plugin_info(self, interface, plugin_name):
        # Find a plugin in the registry and return its info
        # This should remove all plugin loading from the base interfaces and allow us to only load one plugin at a time
    def list_plugins(self, interface):
        # List the plugins available for an interface ONLY based on the registries.
        # This should not load any plugins, just return info from the registries.
```

To use this, I think we would import `PluginRegistry` in `interfaces/base.py`, then attach `plugin_registry` to the base interface class like this:
```
from ___ import PluginRegistry


class BaseInterface:
    plugin_registry = PluginRegistry()
```

Then use the plugin registry to replace calls to `load_all_yaml_plugins` and anywhere that we're using `importlib.resource()`. This will require some refactoring to make appropriate use of `plugin_registry`.

#### Other code locations that this will impact
- This will impact `interfaces/base.py` including:
  - Removing this try/except: https://github.com/NRLMMD-GEOIPS/geoips/pull/240#discussion_r1370851806
  - Refactoring `BaseInterface`, `BaseYamlInterface`, and `BaseModuleInterface`

Other functionality to consider moving to this class (maybe, maybe not):
- `get_entry_point_group()`
- `find_ascii_palette()`
- `find_all_txt_plugins()`
- `find_entry_point()`
- `get_all_entry_points()`
- `list_entry_points()`

### Background and Motivation
Currently we load YAML plugins via the `load_all_yaml_plugins` function. That function gets called every time an interface is instantiated which causes problems in two ways:
1. It causes GeoIPS startup time to be slower than necessary since the function takes about 0.5 seconds every time it is called.
2. It requires us to use a work-around when building the initial registries to avoid a KeyError. This workaround will hide legitimate errors. See: https://github.com/NRLMMD-GEOIPS/geoips/pull/240#discussion_r1370851806

### Writing Unit Tests
I would like to start making it a standard practice to write unit tests for any functions/methods/classes we create. For this issue, I would like to see unit tests that test the following:
- Test the `PluginRegistry` against dummy registry files. These files should include one that is correct (maybe one that was generated by create_registry_plugins) and several that are incorrect. It should ensure that appropriate errors are encountered when it is run against an incorrect registry file. This can be accommodated by adding a keyword argument to `PluginRegistry.__init__()` that allows passing in a list of registry files rather than finding them.
  - When writing these tests, test every function/method/property and attempt to break them using bad registry entries. Raise appropriate errors. Use `pytest.raises` to ensure that the correct errors are being raised in the right circumstances.
- Test the `BaseInterface`, `BaseYamlInterface`, and `BaseModuleInterface` classes and each of their functions/methods/properties. Try to break them in similar ways to above.

### Checklist for Completion
* [ ] Implement the registry class.
* [ ] Write unit tests for the registry class and each of its functions.
* [ ] Instantiate the registry class in the same module where it is defined.
* [ ] Import `PluginRegistry` in `interfaces/base.py` and attach it to the `BaseInterface` class.
* [ ] Refactor each of the interface classes to make appropriate use of the registry.
* [ ] Write unit tests for each of the functions on `BaseInterface`, `BaseYamlInterface`, and `BaseModuleInterface`.
* [ ] Ensure that everything still works.

evrose54:
# Reviewer Checklist

* [x] NO REQUIRED ***updates to other repos*** 

# Related Issues
fixes #389 
fixes #381 

# Testing Instructions
Currently, geoips/tests/integration_tests/full_test.sh is the only test that ensures this is working correctly. Unit tests to be added soon.

# Summary
Now that we have a plugin registry, we should make better use of it. Currently,
create_plugin_registries registered_plugins.yaml outputs are only made use by YAML
plugins. This is inefficient, and should be converted into a class that is used by
every type of interface and plugin. This would also resolve load_all_yaml_plugins
(geoips/geoips_utils.py) being called for each interface, as we instantiate the plugin
registry at the top most level of BaseInterface. This PR will improve the speed of
GeoIPS, and allow for unit testing for the registry itself.

    modified: geoips/geoips/create_plugin_registries.py
    modified: geoips/geoips/geoips_utils.py
    modified: geoips/geoips/interfaces/base.py
    modified: geoips/geoips/interfaces/yaml_based/products.py
    modified: geoips/geoips/interfaces/module_based/filename_formatters.py
    modified: geoips/geoips/plugins/modules/readers/smap_remss_winds_netcdf.py
    added: geoips/geoips/plugin_registry.py
    added: geoips/tests/unit_tests/plugin_registries/test_plugin_registries.py
    added: geoips/tests/unit_tests/plugin_registries/files/*

# Output
Run `create_plugin_registries` or `create_plugin_registries yaml` to see the output of the plugin registry. You can then use IPython and play around with the `PluginRegistry` class to see its functionality.


geoips/plugin_registry.py
```diff
@@ -19,6 +19,121 @@
...
+    def registered_plugins(self):
+        """Find all plugins in registered plugin packages.
+
+        Search the ``registered_plugins.yaml`` of each registered plugin package and add
+        them to the _registered_plugins dictionary
+        """
+        # Load the registries here and return them as a dictionary
+        if not hasattr(self, "_registered_plugins"):
+            self._registered_plugins = {}
+            self._interface_mapping = {}
```
jsolbrig, line 51: Since these attributes are important, it would  be nice to have comments describing what they contain.

jsolbrig, line 51: Is it okay that `self._interface_mapping` isn't set unless `self.registered_plugins` has been called? Should there be a separate property for `self._interface_mapping`? Something like:

```
def set_property_attrs(self):
    # Do what is currently in `registered_plugins`
    # I'm not sure if I like the name `set_property_attrs` so choose whatever you think is appropriate
    # Maybe this function should start with an underscore? I'm not sure.

@property
def registered_plugins(self):
    if not hasattr(self, "_registered_plugins"):
        self.set_property_attrs()
    return self._registered_plugins

@property
def interface_mapping(self):
    if not hasattr(self, "_interface_mapping"):
        self.set_property_attrs()
    return self._interface_mapping

Comment (evrose54) [reply geoips/plugin_registry.py:55]: I'll do this now! And yes, I think this should be a private method.

File: geoips/plugin_registry.py [hunk 2]
```diff
@@ -14,11 +14,207 @@
...
+            validated = interface_module.validator.validate(plugin)
+            return interface_module._plugin_yaml_to_obj(name, validated)
+
+    def list_plugins(self, interface):
+        """List the plugins available for an interface ONLY based on the registries.
+
+        This should not load any plugins, just return info from the registries.
+        """
+        plugin_type = self.identify_plugin_type(interface)
+        return self._registered_plugins[plugin_type][interface]
```
jsolbrig, line 175: Maybe break `list_plugins` into two methods:
- list_interface_plugin_names
- list_interface_plugin_info

geoips/plugin_registry.py
```diff
@@ -14,11 +14,207 @@
...
+                            else:
+                                self.merge_nested_dicts(
+                                    self._registered_plugins[plugin_type][interface],
+                                    interface_dict,
+                                )
+                except TypeError:
+                    raise PluginRegistryError(f"Failed reading {reg_path}.")
+        return self._registered_plugins
+
+    def get_plugin_info(self, interface, plugin_name, just_info=True):
```
jsolbrig, line 97: This function should operate a little differently for YAML-based plugins than for module-based plugins, I think. For module-based plugins it should just return the following:
- plugin type = "module"
- interface
- family
- name
- docstring
- call signature of `call()` (if possible)

For YAML-based plugins, we can be more general, I think, and return all top-level keys that have scalar values (i.e. not mappings or lists). So, for most current plugins we would get:
- plugin_type = "yaml"
- interface
- family
- docstring
But for `product` plugins we would additionally get `source_names`.

geoips/plugin_registry.py
evrose54: I can update those with comments.

geoips/plugin_registry.py
```diff
@@ -0,0 +1,303 @@
...
+            "interpolators",
+            "output_checkers",
+            "output_formatters",
+            "procflows",
+            "readers",
+            "sector_adjusters",
+            "sector_metadata_generators",
+            "sector_spec_generators",
+            "title_formatters",
+        ]
```
mindyls, line 280: Can we avoid explicitly listing these? @evrose54

geoips/plugin_registry.py
```diff
@@ -0,0 +1,294 @@
...
+        yaml_interfaces = [
+            str(info[0]) for info in inspect.getmembers(
+                interfaces.yaml_based,
+                inspect.ismodule)
+        ]
+        module_interfaces = [
+            str(info[0]) for info in inspect.getmembers(
+                interfaces.module_based,
+                inspect.ismodule)
+        ]
```
mindyls, line 271: This is the replacement for listing all of the interfaces explicitly?  @jsolbrig probably has an idea on this as well.

geoips/plugin_registry.py
mindyls: https://github.com/NRLMMD-GEOIPS/geoips/pull/392/files#r1379308126
Address that separately if required. Closing this conversation.

geoips/plugin_registry.py
mindyls: https://github.com/NRLMMD-GEOIPS/geoips/pull/392/files#r1379312623
^ Is that where this was addressed?  Does "get_plugin_info" abstract the process of returning appropriate attributes / fields, and the fields are checked in "validate_plugin_info?

geoips/plugin_registry.py
```diff
@@ -0,0 +1,294 @@
...
+        plugin_type = self.identify_plugin_type(interface)
+        return list(self.registered_plugins[plugin_type][interface].keys())
+
+    def list_plugins(self, interface):
+        """List the plugins available for an interface ONLY based on the registries.
+
+        This should not load any plugins, just return info from the registries.
+        """
+        plugin_type = self.identify_plugin_type(interface)
+        return self.registered_plugins[plugin_type][interface]
```
mindyls, line 150: Are these relevant to @jsolbrig comment https://github.com/NRLMMD-GEOIPS/geoips/pull/392#discussion_r1376758485 ?

At this point, all of these are just accessing the relevant info directly from the existing "registered_plugins" dictionary, so @jsolbrig comment is no longer applicable?

geoips/plugin_registry.py
mindyls: https://github.com/NRLMMD-GEOIPS/geoips/pull/392/files#r1379315304
Does that resolve this conversation?

docs/source/releases/v1_11_7a0.rst
```diff
@@ -256,6 +257,31 @@ require changes to the output products as well.
...
+
+*From issue GEOIPS#389: 2023-10-26, Make better use of the plugin registry*
+
+Now that we have a plugin registry, we should make better use of it. Currently,
+create_plugin_registries registered_plugins.yaml outputs are only made use by YAML
+plugins. This is inefficient, and should be converted into a class that is used by
+every type of interface and plugin. This would also resolve load_all_yaml_plugins
+(geoips/geoips_utils.py) being called for each interface, as we instantiate the plugin
+registry at the top most level of BaseInterface. This PR will improve the speed of
+GeoIPS, and allow for unit testing for the registry itself.
```
mindyls, line 271: I like this explanation!

evrose54: Thanks!

geoips/plugin_registry.py
evrose54: Yes they are. It also expands on it a bit, so we can list all plugins of a given interface, list the names of plugins in a given interface, and also get the information of a plugin specifically. This only reads from the plugin registry.

geoips/plugin_registry.py
```diff
@@ -0,0 +1,294 @@
...
+        for attr in attrs:
+            try:
+                plugin[attr]
+            except KeyError:
+                missing.append(attr)
+        if missing:
+            raise PluginRegistryError(
+                f"Plugin '{name}' is missing the following required "
+                f"top-level properties: '{missing}'"
+            )
```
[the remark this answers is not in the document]
evrose54: Validating the plugin. Information is assembled in 'create_plugin_registries.py'

geoips/create_plugin_registries.py
```diff
@@ -344,10 +370,22 @@ def add_yaml_plugin(filepath, relpath, package, plugins):
...
                         str(subplg_name[1])
                     ] = {
+                        "docstring": yaml_subplg["docstring"],
+                        "family": family,
+                        "interface": interface_module.name,
                         "package": plugin["package"],
+                        "plugin_type": "yaml_based",
+                        "product_defaults": pd,
+                        "source_names": yaml_subplg["source_names"],
                         "relpath": plugin["relpath"],
```
mindyls, line 389: @evrose54  Does this resolve the docstring propagation between product_defualts and products ?

geoips/plugin_registry.py
evrose54: I think all of your comments have been addressed. Let me push up my most recent changes to see if you agree.

geoips/plugin_registry.py
evrose54: These have been created, alongside a specific get_plugin_info function.

geoips/create_plugin_registries.py
```diff
@@ -439,15 +481,25 @@ def add_module_plugin(package, relpath, plugins):
...
+        plugins[interface_name][name] = {
+            "docstring": module.__doc__,
+            "family": module.family,
+            "interface": module.interface,
+            "package": package,
+            "plugin_type": "module_based",
+            "signature": str(signature(module.call)),
+            "relpath": relpath,
+        }
+        del module
```
mindyls, line 502: Here we are acquiring all of the attributes required on a module based plugin, and ensuring everything ends up in the registry?  Again, maybe leave more comments.

evrose54: I can leave more comments. Yes, it is here where we are acquiring all of the required attributes for a module_plugin.

geoips/geoips_utils.py
```diff
@@ -73,19 +74,20 @@ def load_all_yaml_plugins():
...
         if not os.path.exists(pkg_plug_path):
             raise PluginRegistryError(
                 f"Plugin registry {pkg_plug_path} did not exist, "
                 "please run 'create_plugin_registries'"
             )
         # This will include all plugins, including schemas, yaml_based,
         # and module_based plugins.
-        registered_plugins = yaml.safe_load(open(pkg_plug_path, "r"))
+        registered_plugins = pickle.load(open(pkg_plug_path, "rb"))  # nosec
         # Only pull the "yaml_based" plugins here.
```
mindyls, line 91: Should this support both YAML and pickle?

geoips/interfaces/base.py
```diff
@@ -553,6 +563,10 @@ def __repr__(self):
     #         )
     #     return obj
 
+    def __init__(self):
+        """Initialize module plugin interface."""
+        pass
+
     @classmethod
     def _plugin_module_to_obj(cls, name, module, obj_attrs={}):
         """Convert a module plugin to an object.
```
mindyls, line 572: Is this still used ?

mindyls: Oh, yes, in get_plugin.

geoips/create_plugin_registries.py
evrose54: This portion grabs the docstring from the sub-product plugins, or the top level docstring of any other type of yaml plugin. That's why we got an error in data fusion, as some sub-product plugins in 'layered.yaml' did not have docstrings. In general, I think it is appropriate to just use the docstrings of all products, rather than their product defaults, as they are also included in the registries.

geoips/geoips_utils.py
evrose54: I would say no. This is done for efficiency's purpose, and loading yaml takes much longer. Since this is on the backend of GeoIPS, I'd definitely support going the quicker route rather than the longer route.

geoips/create_plugin_registries.py
mindyls: So will this use "products" docstring if it exists, and "product_defaults" docstring if the docstring doesn't exist in the product itself?  In which case we would not have to make any changes to the data_fusion package to get it to work?

evrose54: No, it will just use the docstring from a plugin. I meant that we could find the corresponding docstring in the registry manually, however we can't guarantee that it will be included by the time we process the product plugin. That's why it's important that every product has a docstring, otherwise a KeyError will occur and those products won't be included in the registry.

mindyls: I think there needs to be logic to look in product_defaults for any missing fields in a given product when creating the registry.  So the default docstring would be included in the registry itself when it's created.

geoips/create_plugin_registries.py
mindyls: https://github.com/NRLMMD-GEOIPS/geoips/pull/392/commits/6754848389fd6b02250be244157016c28f81a133

geoips/plugin_registry.py
mindyls: I will let @jsolbrig resolve this one.

mindyls (approved):
About a second to import now, and less than 10 to get all of the product plugins (less time than it used to take to import).  Once all of our conversations are resolved, I am good with this!
```
In [2]: print(datetime.utcnow()); import geoips; print(datetime.utcnow())
2023-11-02 15:12:04.289185
2023-11-02 15:12:05.591306

In [4]: print(datetime.utcnow()); geoips.interfaces.products.get_plugins(); print(datetime.utcnow())
2023-11-02 15:29:13.521275
2023-11-02 15:29:22.130423

In [2]: print(datetime.utcnow()); import geoips; print(datetime.utcnow())
2023-11-02 18:28:20.416648
2023-11-02 18:28:21.225993

In [3]: print(datetime.utcnow()); geoips.interfaces.products.get_plugins(); print(datetime.utcnow())
2023-11-02 18:28:23.198960
2023-11-02 18:28:32.069067
```

evrose54:
I know! Isn't this great?!

geoips/create_plugin_registries.py
evrose54: I modified this to do exactly that in the most recent update to this branch.

mindyls: Indeed you did: https://github.com/NRLMMD-GEOIPS/geoips/pull/392/commits/b30fb53f1dd0985e557fcfb229cec8836ab367be

geoips/plugin_registry.py
mindyls: https://github.com/NRLMMD-GEOIPS/geoips/commit/c48badcd01c9422a65b1b4a579e2d60163096091

There is the commit where it was updated

mindyls: https://github.com/NRLMMD-GEOIPS/geoips/issues/397

Opened an issue to consider moving this to a function on geoips.interfaces, as @jsolbrig thinks this may not be the only place we would want this information.  I'm good resolving this now,  and we can move this functionality later if needed

evrose54: Sounds great! I agree that this may be useful information for other aspects of GeoIPS. I agree that it should probably be moved there in the future.

geoips/geoips_utils.py
mindyls: Replaced with json.  Resolving.

geoips/plugin_registry.py
mindyls: Evan did not include name, because it is the actual key.  Otherwise all points in here have been added to this PR.  Resolving this conversation.

jsolbrig (approved):

PR #401, opened 2023-11-03
377 add log with emphasis function

### LINKED ISSUE #377: Add log with emphasis function
# Requested Update

### Description
Add either LOG.emphasis, or a shared log_with_emphasis function, or LOG.interactive(..., with_emphasis=True)  that will include ***** or other emphasis.
Per:
* https://github.com/NRLMMD-GEOIPS/recenter_tc/pull/28/files#r1361099566
* 

### Background and Motivation

### Alternative Solutions
  
### Environment

### Code to demonstrate issue

### Checklist for Completion
* [ ]

evrose54:
# Reviewer Checklist

# Related Issues
fixes #377 

# Testing Instructions
Run any script that uses output_checkers and you will see the newly formatted output. This can be changed to however you all want it, just tried to create a basic representation that could replace all of the emphasis strings.

# Summary
Scattered throughout GeoIPS (and other packages too), are many Log statements that use
'***********************...' to emphasize the content in between them. While this is
easy to see in the Log output, it's extremely cluttered in the actual code. For this
reason, I've added a 'log_with_emphasis' function to handle this. All you need to do now
is 'log_with_emphasis(LOG.<log_type>, message)', and it will do the rest for you.

    modified: geoips/geoips/commandline/log_setup.py
    modified: geoips/geoips/interfaces/module_based/output_checkers.py
    modified: geoips/geoips/plugins/modules/output_checkers/geotiff.py
    modified: geoips/geoips/plugins/modules/output_checkers/image.py
    modified: geoips/geoips/plugins/modules/output_checkers/netcdf.py
    modified: geoips/geoips/plugins/modules/output_checkers/text.py
    modified: geoips/geoips/plugins/modules/procflows/single_source.py

# Output
I post some output here in a bit.


geoips/commandline/log_setup.py
```diff
@@ -16,6 +16,21 @@
...
+    Parameters
+    ----------
+    LOG: logging function
+        Can be of any type [debug, info, interactive...]
+    message: str
+        The message to be logged with emphasis
+    """
+    LOG("    " + "*" * (len(message) + 8))
+    LOG("    " + "*** " + message + " ***")
+    LOG("    " + "*" * (len(message) + 8))
```
mindyls, line 31: I think we need to pass in a list of strings, and just put one set of asterisks before the block of lines, and another set after.  1.12.0 will have something along these lines coming - so can probably hold off on finishing this until 1.12 is released ?

We probably also just want to have a fixed number of asterisks (not dependent on the length of the message, especially since it can be multiple lines inside the asterisk block)

evrose54:
Don't merge this until v1.12.0-release is merged.

geoips/commandline/log_setup.py
jsolbrig: @evrose54 Can you address Mindy's comment here? I like having the number of asterisks be dynamic, but maybe there should be a maximum length and automatically split `message` into multiple lines to fit within that length?

geoips/commandline/log_setup.py
```diff
@@ -16,6 +16,21 @@
 import sys
```
jsolbrig, line 18: Overall this looks good. I'd like to see one more thing, though. Can you try implementing a unit test for this? It would go in `tests/unit_tests/commandline/log_setup.py` and should make use of pytest's `caplog` feature to make sure the output is what it should be.

I'm asking for this for a few reasons. First, to get in the habit of writing unit tests for everything as we go. Second, to test the `caplog` feature to see if we should apply it elsewhere. Third, because the change asked for in the other conversation will make this more complicated in a way that should likely be tested.

Here's some info on `caplog`: https://pytest-with-eric.com/fixtures/built-in/pytest-caplog/

I think you'd basically just want to test the number of asterisks and number of lines given different length inputs.

jsolbrig: This looks good. Thanks @evrose54!

geoips/commandline/log_setup.py
jsolbrig: This has been fixed.

jsolbrig (approved):

geoips/commandline/log_setup.py
```diff
@@ -16,6 +16,24 @@
...
+    """Log the given message with emphasis provided a certain log type.
+
+    Parameters
+    ----------
+    LOG: logging function
+        Can be of any type [debug, info, interactive...]
+    message: str
+        The message to be logged with emphasis
+    """
+    max_message_len = min(80, max([len(message) for message in messages]))
```
mindyls, line 29: So it will be a max of 80, and shorter if the message is shorter?  That sounds good to me.

geoips/plugins/modules/output_checkers/geotiff.py
```diff
@@ -131,16 +133,17 @@ def outputs_match(plugin, output_product, compare_product):
 
     # subimg_retval = subprocess.call(call_list)
     if retval != 0:
-        LOG.interactive("    *****************************************")
-        LOG.interactive("    *** BAD geotiffs do NOT match exactly ***")
-        LOG.interactive("    ***   output_product: %s ***", output_product)
-        LOG.interactive("    ***   compare_product: %s ***", compare_product)
-        LOG.interactive("    *****************************************")
```
mindyls, line 138: Is there a reason you removed the *** at the beginning of the messages?  I kind of like the ***, do you mind if I put it back in?

geoips/plugins/modules/output_checkers/image.py
```diff
@@ -131,19 +132,14 @@ def outputs_match(plugin, output_product, compare_product, threshold=0.05):
...
-        LOG.interactive(
-            "    ***   compare_product: %s %s ***",
-            np.array(comp_img).shape,
-            compare_product,
-        )
-        LOG.interactive("    ***************************************")
+        message = f"output_product: {np.array(out_img).shape} {output_product}"
+        log_with_emphasis(LOG.interactive, [message])
+        message = f"compare_product: {np.array(comp_img).shape} {compare_product}"
+        log_with_emphasis(LOG.interactive, [message])
```
mindyls, line 142: This should just be one log_with_emphasis call, with all of the lines in a single messages list?

geoips/plugins/modules/output_checkers/geotiff.py
mindyls: That seemed to break black.  I guess the "*** " could also just be added at the beginning of each message line in the log_with_emphasis function?

mindyls: I think we should just add "*** " to the log_with_emphasis functin. Will add comment there.

geoips/plugins/modules/output_checkers/netcdf.py
```diff
@@ -182,6 +193,15 @@ def outputs_match(plugin, output_product, compare_product):
...
+            LOG.interactive,
+            [
+                message,
+                f"output_product: {output_product}",
+                f"compare_product: {compare_product}",
+            ],
+        )
         LOG.interactive("    ****************************************************")
         LOG.interactive(
             "    *** BAD GeoIPS NetCDF files do not match within tolerance *****"
```
mindyls, line 207: Is this going to print the same "Bad GeoIPS NetCDF" message twice?

geoips/plugins/modules/procflows/single_source.py
```diff
@@ -361,17 +361,11 @@ def process_xarray_dict_to_output_format(
...
-    )
+    emphasis_str = "*" * 84
+    LOG.info(f"\n\n{emphasis_str}" f"\n***{print_str}\n{area_def}")
     for key, value in area_def.sector_info.items():
         LOG.info(f"{key}: {value}")
-    LOG.info(
-        "*****************************************************************************"
-        "*******"
-    )
+    LOG.info(emphasis_str)
```
mindyls, line 368: Was this intended to use the log_with_emphasis function also?

geoips/commandline/log_setup.py
```diff
@@ -16,6 +16,24 @@
...
+    LOG("    " + "*" * (max_message_len))
+    for message in messages:
+        LOG("    " + message)
```
mindyls, line 32: Maybe just add "*** " at the beginning of each message?  That would make it match how all the current log with emphasis logs have been printed?

geoips/plugins/modules/output_checkers/image.py
mindyls: Yes, this should be multiple lines passed into a single log_with_emphasis call, please.

geoips/plugins/modules/output_checkers/image.py
```diff
@@ -165,40 +161,38 @@ def outputs_match(plugin, output_product, compare_product, threshold=0.05):
...
     if thresholded_retval != 0:
         bad_inds = np.where(diff_arr != 0)
-        LOG.interactive("    ***************************************")
-        LOG.interactive("    *** BAD Images do NOT match within tolerance ***")
-        LOG.interactive("    *** ***")
-        LOG.interactive(
-            "    *** %s mismatched pixels exceeding threshold %s ***",
-            thresholded_retval,
-            threshold,
+        log_with_emphasis(LOG.interactive, ["BAD Images do NOT match within tolerance"])
```
mindyls, line 164: This line should also be included in the below call to log_with_emphasis, not a separate call, please.

geoips/plugins/modules/output_checkers/netcdf.py
mindyls: I think it is - it looks like the original "LOG.interactive" messages were not removed when the log_with_emphasis call was added?

geoips/plugins/modules/output_checkers/netcdf.py
```diff
@@ -209,14 +229,13 @@ def outputs_match(plugin, output_product, compare_product):
...
-        LOG.info("    ****************************************************************")
+        LOG.info("    " + "*" * 64)
         retval = False
 
     try:
         xarray.testing.assert_identical(compare_xobj, out_xobj)
     except AssertionError as resp:
-        LOG.info("    ****************************************************************")
-        LOG.info("    *** INFORMATIONAL ONLY assert_identical differences *****")
+        log_with_emphasis(LOG.info, ["INFORMATIONAL ONLY assert_identical differences"])
```
mindyls, line 238: This is another one that should have multiple lines passed to a single "log_with_emphasis" call.

geoips/plugins/modules/output_checkers/netcdf.py
```diff
@@ -229,7 +248,7 @@ def outputs_match(plugin, output_product, compare_product):
                 LOG.info(f"    *** maxdiff {varname}: {maxdiff} ***")
             if meandiff != 0:
                 LOG.info(f"    *** meandiff {varname}: {meandiff} ***")
-        LOG.info("    ****************************************************************")
+        LOG.info("    " + "*" * 64)
```
mindyls, line 251: Remove?

mindyls (dismissed):
Please see my comments.  I think we at least want this to produce the same log output as we were producing before.

geoips/commandline/log_setup.py
```diff
@@ -16,6 +16,24 @@
...
 
 
+def log_with_emphasis(LOG, messages):
+    """Log the given message with emphasis provided a certain log type.
+
+    Parameters
+    ----------
+    LOG: logging function
+        Can be of any type [debug, info, interactive...]
+    message: str
```
biosafetylvl5, line 26:
```python
>>> import geoips.commandline.log_setup as log_setup
>>> import logging
>>> LOG = logging.getLogger("fake_name")
>>> log_setup.log_with_emphasis(LOG.critical, ["A very long string just for testing this. "*5])
```

Outputs
```
    ********************************************************************************
    A very long string just for testing this. A very long string just for testing this. A very long string just for testing this. A very long string just for testing this. A very long string just for testing this. 
    ********************************************************************************
```

biosafetylvl5: but 
```python
>>> log_setup.log_with_emphasis(LOG.critical, "test")
```

Outputs
```
    *
    t
    e
    s
    t
    *
```

biosafetylvl5: Is this intended? Am I testing this correctly?

biosafetylvl5: If I *am* testing correctly, please fix. Else please update docstring with an example to help me/future folks out 🙂

biosafetylvl5 (changes requested):
I found a bug/hole in the documentation, and a place where this needs to be improved before it can be merged. See my comments :)

geoips/commandline/log_setup.py
jsolbrig: I can see a couple of different ways that this could work. This could accept a single string and word-wrap it, regardless of its contents. It could also accept multiple strings and do something intelligent with them like treat them as paragraphs to be wrapped separately. I'd lean towards the simple option just for KISS purposes unless there is a reason we want paragraphs.

With single strings, @biosafetylvl5's first example would just be this:
```
>>> import geoips.commandline.log_setup as log_setup
>>> import logging
>>> LOG = logging.getLogger("fake_name")
>>> log_setup.log_with_emphasis(LOG.critical, " ".join(["A very long string just for testing this. "*5]))
```

and their `"test"` example would work as-is:

```
>>> log_setup.log_with_emphasis(LOG.critical, "test")
*********
** test **
*********
```

Alternatively, if we allowed multiple messages in a single call, it could consider them to be paragraphs:

```
def log_with_emphasis(LOG, *messages):
    ...
```

And the call would look like this:
```
log_with_emphasis(LOG.critical, "A very long string just for testing this.", "A very long string just for testing this.", ...)
```

Which would make something like this:
```
***********************************************
** A very long string just for testing this. **
**                                           **
** A very long string just for testing this. **
**                                           **
** A very long string just for testing this. **
***********************************************
```

geoips/commandline/log_setup.py
jsolbrig: `os.get_terminal_size()` could be used to retrieve the shape of the current terminal.

mindyls: I think we are way overthinking this.  If someone passes in a long line, just print it long.

If someone cares about how long their lines are, they can just adjust themselves.  I wouldn't worry at all about wrapping the individual lines.

If the line extends beyond the 80 characters, the ***** will stil just be 80 characters, and the line will extend beyond the ****s

geoips/commandline/log_setup.py
```diff
@@ -16,6 +16,24 @@
 import sys
 
 
+def log_with_emphasis(LOG, messages):
+    """Log the given message with emphasis provided a certain log type.
```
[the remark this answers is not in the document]
mindyls: I think the person passing in the lines should just be responsible for making sure each line is the length they want it...

Pass in multiple lines of the length you want them printed to the screen.  Normal calls to LOG or print don't modify the length of your lines, I think we can safely assume people pass line lengths they want printed.

geoips/commandline/log_setup.py
mindyls: Oh, I was expecting this to be a list of strings, with a newline between each string, and the ***s before and after the lines.

We could just check if it is a list in the log_with_emphasis function, and print a single line if not, else loop through the strings in the list if it is a list ?

biosafetylvl5: I like what you're thinking @mindyls! I think you're right about KISS, so maybe we should just word-wrap single strings? If you have a list of strings you can always call the log function multiple times.

eg.

```python
for string in ["A very long string just for testing this.", "A very long string just for testing this.", ...]:
    log_with_emphasis(LOG.critical, string)
```

Output:
```
***********************************************
** A very long string just for testing this. **
***********************************************
***********************************************
** A very long string just for testing this. **
***********************************************
***********************************************
***********************************************
** A very long string just for testing this. **
***********************************************
... (etc.)
```

geoips/commandline/log_setup.py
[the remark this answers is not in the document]
biosafetylvl5: I agree with @mindyls  here - that the convention is to make the user handle the string length for logging functions so we probably should stick with that and not auto-wrap the logged string.

I'm concerned because this function "wraps" the log string in `*`s which means that if you have a long string you could get:

```
***********************************************
** Testing strings for various purposes can be quite intriguing, as it involves exploring different functionalities and edge cases. Testing strings for various purposes can be quite intriguing, as it involves exploring different functionalities and edge cases. Testing strings for various purposes can be quite intriguing, as it involves exploring different functionalities and edge cases. **
***********************************************
```

which wraps like this on my terminal:

image

which is a bit confusing. I feel like this:

```
***********************************************
** A very long string just for testing this. **
** A very long string just for testing this. **
** A very long string just for testing this. **
***********************************************
```
is a lot more readable, but requires us to wrap the string which breaks convention on logging not handling the wrapping of the output.

At this point I'm wondering if simply pre-pending the logged line with `***` might be cleaner and more in convention? What is the purpose of this new logging level?

eg. the output would be:

```
*** Testing strings for various purposes can be quite intriguing, as it involves exploring different functionalities and edge cases. Testing strings for various purposes can be quite intriguing, as it involves exploring different functionalities and edge cases. Testing strings for various purposes can be quite intriguing, as it involves exploring different functionalities and edge cases.
```

What do you think @mindyls ?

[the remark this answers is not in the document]
biosafetylvl5: Just an idea - would this be better with colorized output? (eg. we could do a different color for emphasis?)

geoips/commandline/log_setup.py
biosafetylvl5: So the intended output would be:

```
********************************************************************************
*** Testing strings for various purposes can be quite intriguing, as it involves exploring different functionalities and edge cases. Testing strings for various purposes can be quite intriguing, as it involves exploring different functionalities and edge cases. Testing strings for various purposes can be quite intriguing, as it involves exploring different functionalities and edge cases. ***
********************************************************************************
```

This wraps like this on a small terminal:

image

and this on a larger terminal:

image

And with a smaller log message, the intended output would be:

```
********************************************************************************
*** Testing string. ***
********************************************************************************
```

This wraps like this on a small terminal:

image

and this on a larger terminal:

image

Is that right?

geoips/commandline/log_setup.py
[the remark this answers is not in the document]
mindyls: The purpose of this new logging level is that we wanted to consolidate all of the "*********" scattered throughout the code, so wanted to replace it with a new log level.

The only requirements from my perspective are to replicate what those "*****"s did before:
* pass in a list of strings
* Put "*"*80 before the strings
* print the strings one at a time
* Put "*"*80 after the strings.

geoips/commandline/log_setup.py
mindyls: I would rather just have it replicate what it did before

* pass in a list of strings
* Put "*"*80 before the strings
* print the strings one at a time
* Put "*"*80 after the strings.

If we wanted to allow passing an individual string as well (to avoid your 
t
e
s
t
above), we could just check in log_with_emphasis if a list or string was passed in.

geoips/commandline/log_setup.py
biosafetylvl5: My mistake - I forgot it'll shorten the `*`s if the message is shorter.

biosafetylvl5:
Okay to summarize all the comments above:

1. The goal of this PR is to remove statements like `print("*"*80)` from the code base which were used to emphasize some logged output
2. It is NOT intended to improve/change the logging, or change the way these emphasized messages are output to the console
3. The logging function should be able to take in a list or a string. If the input is a list, log out `"\n".join(list)`
4. Python logging functions do not wrap their output or do extra formatting on their logged messages. However, the output of the log with emphasis function's output suggests wrapping could be helpful. Regardless of whether or not this is the ideal output, this PR does *not* aim to address this.

The current status of this PR is:

1. A new function that logs with emphasis and relevant documentation were added
2. The function accepts lists. We suggested above it should also accept strings.
3. Though much of the code that *should* be replaced with the new function have been replaced, there remain some parts that still need updating. @mindyls commented on a number of these.

My personal opinion is that:

1. The logging function should be modified to accept strings
2. The rest of the code should be edited to remove straggler `"*"*80`s
3. A new issue should be opened about the specific format of the logging with emphasis output
4. This PR should be merged

This plan allows us to immediately remove the `"*"*80` parts of the code, and refactors the logging functionality. This is worth merging as it's an incremental improvement to the code. Opening a new issue allows us to further improve/discuss the exact output of the logging function separate from the refactoring while not holding this PR from merging.

How does that sound @mindyls, @jsolbrig?

evrose54:
I've modified the output to look like this. What do you think?

```
In [10]: log_with_emphasis(<logger_instance>, "hello", "howdy", "what's up")
***************
** hello     **
** howdy     **
** what's up **
***************

In [11]: log_with_emphasis(<logger_instance>, "hello", "howdy", "what's up", "this is a very long string that we are not going to match at the top or bottom because it really is too long ok I think this is long enough")
**************************************************************************************
** hello                                                                            **
** howdy                                                                            **
** what's up                                                                        **
** this is a very long string that we are not going to match at the top or bottom because it really is too long ok I think this is long enough **
**************************************************************************************
```

jsolbrig:
@evrose54 I think I'd like to use the textwrap package to wrap lines that are longer than the maximum width so that everything would fit inside the box. Something like this:

```
In [11]: log_with_emphasis(<logger_instance>, "hello", "howdy", "what's up", "this is a very long string that we are not going to match at the top or bottom because it really is too long ok I think this is long enough")

************************************************************************************
** hello                                                                          **
** howdy                                                                          **
** what's up                                                                      **
** this is a very long string that we are not going to match at the top or bottom **
** because it really is too long ok I think this is long enough                   **
************************************************************************************
```

Where the width of the box would be determined by the longest remaining line after wrapping the text. If your maximum width were 80 characters, I think you can do this by just calling `wrapped_lines = textwrap.wrap(line, width=80)`.

biosafetylvl5:
This is much prettier! I think @mindyls point that this breaks the logging standard set by python is a valid one. If we want to discuss the specifics of the formatting on this PR I think we should consider the implications of breaking the standard.

If I'm looking through lots of geoips logs and want to pull out all the emphasized logs, I might do something like:

```bash
cat geoips_logs.log | grep "\*\*"
```

which using @evrose54s example, we'd get something like:

```
************************************************************************************
** hello                                                                          **
** howdy                                                                          **
** what's up                                                                      **
** this is a very long string that we are not going to match at the top or bottom **
** because it really is too long ok I think this is long enough                   **
************************************************************************************
************************************************************************************
** hello                                                                          **
** howdy                                                                          **
** what's up                                                                      **
** this is a very long string that we are not going to match at the top or bottom **
** because it really is too long ok I think this is long enough                   **
************************************************************************************
************************************************************************************
** hello                                                                          **
** howdy                                                                          **
** what's up                                                                      **
** this is a very long string that we are not going to match at the top or bottom **
** because it really is too long ok I think this is long enough                   **
************************************************************************************
```

## Example 1: Counting outputs

If I wanted to do something with these, the extra line breaks would be a complicating issue. If I was counting "WARNING" lines, I could `grep -f geoips_logs.log -c "WARNING"`. However, if I wanted to count the number of emphasized log messages, I - the end user - would need to do something like `$(('grep -f geoips_logs.log -c "*\*\*\*\*" ' / 2))` which feels more like a workaround. 


## Example 2: Log processing with python

If I wanted to loop over every "WARNING" log file message I would need to:

```python
for open("geoips_logs.log", "r") as f:
    warning_messages = list(filter(lambda line: "WARNING" in line, f.readlines()))
```

If I wanted to loop over the emphasized log outputs, I'd need to do something more state-based. Eg:

```python
for open("geoips_logs.log", "r") as f:
    in_emphasis_log = False
    emphasized_messages = []
    current_message_lines = []
    for line in f.readlines():
        if "\*\*\*" in line and (other_checks): # other checks so we don't catch other logged output
             if in_emphasis_log:
                 emphasized_messages.append(" ".join(current_message_lines))
                 current_message_lines = []
             in_emphasis_log = not in_emphasis_log
        elif in_emphasis_log:
             current_message_lines.append(
                       line.replace("\*\* ", "").strip()
              )
```

which is both harder to follow, more compute intensive and difficult to parallelize.

## Proposed fix

If the boxed output in the user-facing CLI are the ideal output, we could consider using different logging formatters for files vs the CLI ([example](https://stackoverflow.com/a/60473609)). This would allow us to break conventions for standard python loggers *but* not face the technical debt of more convoluted logging outputs. 

Eg. users would see
```
************************************************************************************
** hello                                                                          **
** howdy                                                                          **
** what's up                                                                      **
** this is a very long string that we are not going to match at the top or bottom **
** because it really is too long ok I think this is long enough                   **
************************************************************************************
```

in the command line, but in any logged out files the output would follow pythons conventions. Eg:

```
INFO :: EMPHASIS :: hello howdy what's up this is a very long string that we are not going to match at the top or bottom because it really is too long ok I think this is long enough
```

## Alternative

We decide it is unlikely we'll run into issues like this in the future or that discussing this is unproductive and take on the technical debt.

evrose54:
Hopefully this works with everyone now. @jsolbrig I made ``log_with_emphasis`` make use of ``textwrap`` as you suggested. That was a good library to use for our needs. Thanks for the suggestion. Shown below is example output of what ``log_with_emphasis`` will produce. Lines are a maximum of 80 characters in length, including the pre/post-fixes that have been added to intermediary lines.

```
In [4]: log_with_emphasis(print, "hello", "howdy", "what's up", "this is a very long string that we are not going to match at the top or bottom because it really is too long ok I think this is long enough")
*****************************************************************************
** hello                                                                   **
** howdy                                                                   **
** what's up                                                               **
** this is a very long string that we are not going to match at the top or **
** bottom because it really is too long ok I think this is long enough     **
*****************************************************************************
```

biosafetylvl5:
Looks great @evrose54 ! I think we just need to update the rest of the examples that mindy highlighted that still use direct `print("*"*80)` statements and then we can merge

biosafetylvl5:
Looks good to me!

geoips/plugins/modules/output_checkers/image.py
biosafetylvl5: Fixed

docs/source/releases/v1_12_2a0.rst
```diff
@@ -13,16 +13,44 @@
...
+::
+
+    modified: geoips/geoips/commandline/log_setup.py
+    modified: geoips/geoips/interfaces/module_based/output_checkers.py
+    modified: geoips/geoips/plugins/modules/output_checkers/geotiff.py
+    modified: geoips/geoips/plugins/modules/output_checkers/image.py
+    modified: geoips/geoips/plugins/modules/output_checkers/netcdf.py
+    modified: geoips/geoips/plugins/modules/output_checkers/text.py
+    modified: geoips/geoips/plugins/modules/procflows/single_source.py
+    added: geoips/tests/unit_tests/commandline/log_setup.py
```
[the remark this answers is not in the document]
biosafetylvl5: Fixed

biosafetylvl5 (approved):

evrose54:
Just fixed the conflict with the release notes. Once Mindy approves this PR, it should be good to merge in.

biosafetylvl5:
I'm updating the plugin repos:

- [ ] geoips_clavrx
- [ ] recenter_tc
- [ ] data_fusion
- [ ] geoips_plugin_example
- [ ] geocolor
- [ ] true_color
- [ ] synth_green
- [ ] rayleigh
- [ ] lunarref
- [ ] fortran_utils
- [ ] ancildat
- [ ] ryglickicane
- [ ] geoips_tropics
- [ ] overcast_package
- [ ] template_fusion_plugin
- [ ] geoips_awips
- [ ] akima86

jsolbrig (approved):
I'm good with this PR now. Thank you @evrose54 for putting this together and @biosafetylvl5 for your thorough review!

mindyls (approved):
Looks good,  thanks for the thorough discussion and review!

PR #417, opened 2023-12-21
416 replace sys argv with argparse

### LINKED ISSUE #416: Replace create_plugin_registries 'sys.argv' calls with 'argparse' library
# Requested Update

### Description
Currently, create_plugin_registries.py uses sys.argv to parse arguments for this program. We should avoid using this in all of GeoIPS, as it follow the conventions that will be upheld by the new CLI. Argparse is more dynamic, and can provide useful feedback to the user for how to use a certain program.

### Background and Motivation
Lots of GeoIPS (and the eventual CLI) uses argparse, and we should adhere to this convention for all GeoIPS code.

### Alternative Solutions
Leave the code as is and deal with it. 
  
### Environment

### Code to demonstrate issue
geoips/geoips/create_plugin_registries.py

### Checklist for Completion
* [ ]  Change sys.argv calls to argparse calls.

evrose54:
Reviewer Checklist

# Related Issues
fixes #416 

# Summary
Currently create_plugin_registries.py uses sys.argv calls to generate its arguments
rather than argparse. This doesn't follow GeoIPS conventions nor that of the CLI that
will soon be updated. We should replace these calls with argparse architecture, so that
this code follows current conventions, as does other GeoIPS code.


docs/source/releases/v1_11_7a0.rst
```diff
@@ -32,6 +32,7 @@ Version 1.11.7a0 (2023-10-26)
 * Added unit test configuration file
```
mindyls, line 32: Opening conversation to remind us to NOT merge this until v1.12.0-release has been merged to main, and this destination branch is automatically updated to point to main.

jsolbrig (approved):

geoips/create_plugin_registries.py
```diff
@@ -848,9 +849,30 @@ def main():
...
+        description=description,
+    )
+    argparser.add_argument(
+        "-s",
+        "--save_type",
+        type=str.lower,
+        default="json",
+        choices=["json", "yaml"],
+        help="Format to write registries to. This will also be the file extension.",
+    )
```
mindyls, line 872: There is an incoming change that allows passing a package name into create_plugin_registries - we will update this to allow a "-p" option when 1.12.1 is released.

mindyls (approved):
Looks good to me.  Note my comment about incoming package argument (allowing calling create_plugin_registires with only a single package).  Should be an easy update.

geoips/create_plugin_registries.py
jsolbrig: Sounds good. I think we can handle this when 1.12.1 comes through. I'm going to resolve and merge.

jsolbrig (approved):

Issue #426, opened 2024-02-01
GeoIPS will not run when installed via `pip install .` without the "test" target


jsolbrig:
# Requested Update

### Description
GeoIPS currently fails to run if installed by simply calling `pip install .`. It must, instead, be installed with `pip install .[test]`. This is because `pytest` is used in `geoips/plugin_registry.py` but is not included in the base GeoIPS installation requirements.

There are three reasonable solutions to this issue:
1. Include pytest in the requirements for base GeoIPS.
2. Only import pytest if it is available and raise an error if it is encountered but not imported.
3. Remove pytest from plugin_registry.py and move the tests into the unit_test directory.

Personally, I would prefer to keep unit tests out of the actual code base, but maybe there is a reason that I'm unaware of that requires the use of `pytest.xfail` in `plugin_registry.py` rather than in the external unit tests.

### Code to demonstrate issue
- In a fresh environment, install GeoIPS via `pip install .`
- Try calling `run_procflow -h`

You will see something like:
```
geoips_user@eae1915cc33f:/geoips$ cd ..
geoips_user@eae1915cc33f:/$ run_procflow -h
Traceback (most recent call last):
  File "/home/geoips_user/.local/bin/run_procflow", line 5, in <module>
    from geoips.commandline.run_procflow import main
  File "/home/geoips_user/.local/lib/python3.10/site-packages/geoips/__init__.py", line 31, in <module>
    from geoips import interfaces
  File "/home/geoips_user/.local/lib/python3.10/site-packages/geoips/interfaces/__init__.py", line 15, in <module>
    from geoips.interfaces.module_based.algorithms import algorithms
  File "/home/geoips_user/.local/lib/python3.10/site-packages/geoips/interfaces/module_based/algorithms.py", line 15, in <module>
    from geoips.interfaces.base import BaseModuleInterface
  File "/home/geoips_user/.local/lib/python3.10/site-packages/geoips/interfaces/base.py", line 297, in <module>
    class BaseInterface:
  File "/home/geoips_user/.local/lib/python3.10/site-packages/geoips/interfaces/base.py", line 309, in BaseInterface
    from geoips.plugin_registry import plugin_registry
  File "/home/geoips_user/.local/lib/python3.10/site-packages/geoips/plugin_registry.py", line 21, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'
```

### Checklist for Completion
* [ ]

evrose54:
[...]

PR #442, opened 2024-02-16
Move PyTest functionality from PluginRegistry Class to Unit Tests

evrose54:
# Reviewer Checklist

# Related Issues
fixes NRLMMD-GEOIPS/geoips#426

# Testing Instructions
To run the new unit tests, simply run ``pytest test_plugin_registries.py -v``. This will unit test the plugin registry and give you verbose output of what occurs. All passes should test other than three, which fail (expected failure). This file can be found at geoips/tests/unit_tests/plugin_registries/test_plugin_registries.py.

# Summary
*From GEOIPS#426: 2024-02-02, GeoIPS won't run pip install . w/out [test] target*

GeoIPS currently fails to run if installed by simply calling ``pip install .``. It must,
instead, be installed with pip install .[test]. This is because pytest is used in
geoips/plugin_registry.py but is not included in the base GeoIPS installation
requirements.

To remediate this, we've since moved the pytest validation functionality off of the
PluginRegistry class itself, and moved it into unit tests at
``tests/unit_tests/plugin_registries/test_plugin_registries.py``, under class
PluginRegistryValidator. This removes the need for PyTest to be a GeoIPS dependency,
and is now a Unit Test dependency. If users want to validate their plugin registry, they
will have to manually install pytest to validate from now on.

    - modified: geoips/plugin_registry.py
    - modified: tests/unit_tests/plugin_registries/test_plugin_registries.py

# Output
Shown below is the expected result of running the pytest command listed above.

Screenshot 2024-02-16 at 11 41 01 AM


jsolbrig:
@evrose54 I think I'd like to talk about this PR a little bit. You did what I asked, but I think we should think about how the tests are structured a little more carefully and whether some of the validation should remain on in plugin_registry.py.

Questions:
- Should some of the validation remain in the plugin registry code?
- Can we just remove the pytest calls from any tests that we move back?
- Can we move a lot of the validation functions into the `test_` functions rather than defining two classes here?

tests/unit_tests/plugin_registries/test_plugin_registries.py
```diff
@@ -10,27 +10,214 @@
...
+class TestPluginRegistry:
+    """
+    Pytest-based Unit Test for the PluginRegistry Class.
+
+    Note: Since we are not able to initlialize this class due to restrictions placed by
+    Pytest, if you want to change the test files used, simply replace the location below
+    with the location of your new test files.
+    """
+
+    _default_fpaths = glob(
```
biosafetylvl5, line 208: why is this marked as a pseudo private var?

tests/unit_tests/plugin_registries/test_plugin_registries.py
```diff
@@ -10,27 +10,214 @@
...
         str(__file__).replace("test_plugin_registries.py", "files/**/*.yaml"),
         recursive=True,
     )
 
-    plugin_registry = PluginRegistry(fpaths)
+    _pr_validator = PluginRegistryValidator(_default_fpaths)
+
+    # Couldn't implement this class via inheritance because PyTest Raised this error:
```
biosafetylvl5, line 215: If including a comment like this, we should explain what the error means, and why it's fine

tests/unit_tests/plugin_registries/test_plugin_registries.py
```diff
@@ -10,27 +10,214 @@
...
-class TestPluginRegistry:
+class PluginRegistryValidator(PluginRegistry):
     """Subclass of PluginRegistry which adds functionality for unit testing."""
 
-    fpaths = glob(
+    def __init__(self, fpaths=None):
+        """Initialize TestPluginRegistry Class."""
+        super().__init__(fpaths)
+
+    def validate_plugin_types_exist(self, reg_dict, reg_path):
```
biosafetylvl5, line 37: this is a direct copy-paste - no edits

docs/source/releases/v1_12_2a0.rst
```diff
@@ -0,0 +1,42 @@
...
+instead, be installed with pip install .[test]. This is because pytest is used in
+geoips/plugin_registry.py but is not included in the base GeoIPS installation
+requirements.
+
+To remediate this, we've since moved the pytest validation functionality off of the
+PluginRegistry class itself, and moved it into unit tests at
+``tests/unit_tests/plugin_registries/test_plugin_registries.py``, under class
+PluginRegistryValidator. This removes the need for PyTest to be a GeoIPS dependency,
+and is now a Unit Test dependency. If users want to validate their plugin registry, they
+will have to manually install pytest to validate from now on.
```
biosafetylvl5, line 37: Your description of why this is needed on slack was very helpful - can you integrate it into the docs somewhere?

```
Just to ensure it's in the correct state. This was initially automatically done when it was created, but since it invoked pytest and that's not a required dependency, I had to move this to a unit-test module instead. All in all, the user doesn't really need to validate their own registry. It should not be created if an error occurs (duplicate plugin names under the same interface, poorly formatted plugin, etc.), however as developers Jeremy wanted the ability to unit test the PluginRegistry to make sure everything was being done as expected.
```

biosafetylvl5 (changes requested):
This is mostly a copy-paste move, and as such doesn't have much new code to review. I asked for some documentation. @jsolbrig I think you should weigh in on whether this is a good implementation. Let me know if you'd like me to review the code more in-depth.

tests/unit_tests/plugin_registries/test_plugin_registries.py
jsolbrig: I think Evan is saying that he isn't subclassing in order to avoid having this error occur. Rather than inheriting from `PluginRegistry`, he is setting it to the class attribute `plugin_registry`.

docs/source/releases/v1_12_2a0.rst
evrose54: I can add this to the documentation.

tests/unit_tests/plugin_registries/test_plugin_registries.py
evrose54: Changed it to ``default_fpaths`` instead.

tests/unit_tests/plugin_registries/test_plugin_registries.py
evrose54: Added a comment explaining this more.

tests/unit_tests/plugin_registries/test_plugin_registries.py
biosafetylvl5: Thanks! :)

docs/source/releases/v1_12_2a0.rst
biosafetylvl5: This is wonderful - thank you so much!

biosafetylvl5 (approved):
Thank you for making the changes I requested @evrose54 ! :) I'll leave it up to @jsolbrig to review the rest.

jsolbrig (approved):

tests/unit_tests/plugin_registries/test_plugin_registries.py
```diff
@@ -10,27 +10,217 @@
...
```
mindyls, line 37: So from validate_plugin_types through validate_registry_interfaces, this is all a direct copy/paste from geoips/plugin_registry.py with no additional udpates?

These methods were added to a new "PluginRegistryValidator" class in the test_plugin_registries.py test module (with an init added)?

tests/unit_tests/plugin_registries/test_plugin_registries.py
```diff
@@ -10,27 +10,217 @@
...
+class TestPluginRegistry:
+    """
+    Pytest-based Unit Test for the PluginRegistry Class.
+
+    Note: Since we are not able to initlialize this class due to restrictions placed by
+    Pytest, if you want to change the test files used, simply replace the location below
+    with the location of your new test files.
+    """
+
+    default_fpaths = glob(
```
mindyls, line 208: This is the original existing TestPluginRegistry class (it is just showing up as new right here, because git got confused and showed the class TestPluginRegistry as deleted above)?

tests/unit_tests/plugin_registries/test_plugin_registries.py
```diff
@@ -41,25 +231,25 @@ def generate_id(fpath):
...
-        print(self.plugin_registry.registered_plugins)
-        assert isinstance(self.plugin_registry.registered_plugins, dict)
-        assert "yaml_based" in self.plugin_registry.registered_plugins
-        assert "module_based" in self.plugin_registry.registered_plugins
-        assert "text_based" in self.plugin_registry.registered_plugins
+        print(self.pr_validator.registered_plugins)
+        assert isinstance(self.pr_validator.registered_plugins, dict)
+        assert "yaml_based" in self.pr_validator.registered_plugins
+        assert "module_based" in self.pr_validator.registered_plugins
+        assert "text_based" in self.pr_validator.registered_plugins
```
mindyls, line 238: This is just updated so now it uses the internal PR Validator class rather than the imported plugin_registry validator form geoips/plugin_registry.py, right?

mindyls (approved):
Just resolve my comments if they are correct - I was just leaving notes to make sure I understood what was being updated correctly.

Looks good, and I got 0 return.

tests/unit_tests/plugin_registries/test_plugin_registries.py
jsolbrig: Yes, this is right. We discussed leaving them where they were and extracting the pytest portions, but determined that these tests really are only needed at development time and aren't needed at run time so it made sense to move them to the unit tests.

tests/unit_tests/plugin_registries/test_plugin_registries.py
jsolbrig: Yep, that's right. It thought the previous part was a renaming rather than a move.

tests/unit_tests/plugin_registries/test_plugin_registries.py
jsolbrig: Correct. The the PluginRegistry class doesn't really do any validation anymore. It has all been moved here.

Now that I think about it, though, we should test what happens now if the plugin registry is corrupted or malformed. Are the resulting errors helpful? If not, can we make them helpful?

@evrose54 What do you think on this?

evrose54: That's probably a good idea. I don't love that the Plugin Registry isn't validated during it's creation, but it will check to make sure that every plugin added is not duplicately named and that every plugin added is valid in its nature. The way I see it we have two options:

1. Run the validation routines when the plugin registry is created, and remove references to pytest
2. Keep the validation routines as a unit test, however run some final validation at the end of the registry's creation to ensure it's in a correct state. 

A side note:
When we load in the registry to use in GeoIPS, via get_plugin, ... it might be smart to perform some validation of the contents it's reading. Users could possible modify the registry files manually and we should check that the registries are in fact valid when loading them in. Currently we're assuming that after they've been created they are valid, and do no validation during runtime.

jsolbrig: I'm debating whether I think this is safe. I'm not sure if I'm comfortable only validating at creation time, but maybe it is okay.

Maybe another option. Would it be possible to separate the validation routines from the pytest calls? Keep some or all of the validation in the original class, but have it raise errors rather than call pytest.xfail, then catch those errors in the unit tests?

I think that I'm good with approving this PR, but I think we should create another issue to think about where validation should actually occur and how to handle unit testing for this. For example, maybe the part that should be getting tested in the unit tests is actually the validation functions themselves.

evrose54: That's a fair point. I'll create an issue that we should discuss where the validation should occur, and what the unit tests should actually test. Once that is done, I'll merge this PR.

evrose54: Going to merge now, see issue #496 for more information.

PR #464, opened 2024-04-01
Finish log with emphasis function

biosafetylvl5:
See #377 for more information. This corrects the release notes and converts some straggler calls that *should* have been replaced with `log_with_emphasis`.

Passes the same unit tests as `main` as well as the integration tests.

```
Package: geoips_base
Total run time: 62 seconds
Number data types run: 3
Number data types failed: 0
Mon Apr  1 20:38:17 UTC 2024
```


geoips/plugins/modules/output_checkers/netcdf.py
```diff
@@ -204,34 +203,32 @@ def outputs_match(plugin, output_product, compare_product):
...
-                LOG.interactive(f"    *** mindiff {varname}: {mindiff} ***")
+                log_with_emphasis(LOG.interactive, "mindiff {varname}: {mindiff}")
                 diffout += [f"mindiff {varname}: {mindiff}\n"]
             if maxdiff != 0:
-                LOG.interactive(f"    *** maxdiff {varname}: {maxdiff} ***")
+                log_with_emphasis(LOG.interactive, "mindiff {varname}: {maxdiff}")
                 diffout += [f"maxdiff {varname}: {maxdiff}\n"]
             if meandiff != 0:
-                LOG.interactive(f"    *** meandiff {varname}: {meandiff} ***")
+                log_with_emphasis(LOG.interactive, "mindiff {varname}: {meandiff}")
```
jsolbrig, line 212: Maybe it would be better for these to be done in a single call to log_with_emphasis to avoid too many sets of asterisk boxes. Alternatively, maybe log_with_emphasis should have an option to turn off the leading and trailing lines.

geoips/plugins/modules/output_checkers/netcdf.py
```diff
@@ -204,34 +203,32 @@ def outputs_match(plugin, output_product, compare_product):
...
             if mindiff != 0:
-                LOG.info(f"    *** mindiff {varname}: {mindiff} ***")
+                log_with_emphasis(LOG.info, "mindiff {varname}: {mindiff}")
             if maxdiff != 0:
-                LOG.info(f"    *** maxdiff {varname}: {maxdiff} ***")
+                log_with_emphasis(LOG.info, "mindiff {varname}: {maxdiff}")
             if meandiff != 0:
-                LOG.info(f"    *** meandiff {varname}: {meandiff} ***")
-        LOG.info("    " + "*" * 64)
+                log_with_emphasis(LOG.info, "mindiff {varname}: {meandiff}")
```
jsolbrig, line 231: Combine these lines into a single call to log_with_emphasis.

jsolbrig (changes requested):
I left comments in two spots where the log_with_emphasis calls should be combined into an single call. The output right now would look like this:

```
************
** line 1 **
************
************
** line 2 **
************
************
** line 3 **
************
```

but should look like this:
```
************
** line 1 **
** line 2 **
** line 3 **
************
```

biosafetylvl5:
Good call @jsolbrig - thank you for revewing my code and offering feedback! I've added some code to this PR to fix the logged output so it's more 

```

************
** line 1 **
** line 2 **
** line 3 **
************
```

than

```
************
** line 1 **
************
************
** line 2 **
************
************
** line 3 **
************
```

I also refactored the code since it was duplicated

geoips/plugins/modules/output_checkers/netcdf.py
```diff
@@ -241,6 +220,64 @@ def outputs_match(plugin, output_product, compare_product):
...
+    meandiff = (object2[compare_key] - object1[compare_key]).mean()
+    messages = []
+    if mindiff != 0:
+        messages.append(f"mindiff {compare_key}: {mindiff}\n")
+    if maxdiff != 0:
+        messages.append(f"maxdiff {compare_key}: {maxdiff}\n")
+    if meandiff != 0:
+        messages.append(f"meandiff {compare_key}: {meandiff}\n")
+    log_with_emphasis(LOG.info, messages)
+    return messages
```
jsolbrig, line 278: Testing this function lead me to uncover a bug in `log_with_emphasis`. If `log_with_emphasis` is called with a zero-length string as any of its message arguments it raises the following error. I'd prefer that it is able to handle zero-length strings and just represent them as an empty line.

```
  File "/local/home/jsolbrig/geoips/github/geoips/tests/scripts/test_log_with_emphasis.py", line 5, in <module>
    log_with_emphasis(log.info, "")
  File "/local/home/jsolbrig/geoips/github/geoips/geoips/commandline/log_setup.py", line 39, in log_with_emphasis
    max_message_len = min(74, max([len(wmessage) for wmessage in wrapped_messages]))
ValueError: max() arg is an empty sequence
```

geoips/plugins/modules/output_checkers/netcdf.py
```diff
@@ -241,6 +220,64 @@ def outputs_match(plugin, output_product, compare_product):
...
+    """
+    maxdiff = (object2[compare_key] - object1[compare_key]).max()
+    mindiff = (object2[compare_key] - object1[compare_key]).min()
+    meandiff = (object2[compare_key] - object1[compare_key]).mean()
+    messages = []
+    if mindiff != 0:
+        messages.append(f"mindiff {compare_key}: {mindiff}\n")
+    if maxdiff != 0:
+        messages.append(f"maxdiff {compare_key}: {maxdiff}\n")
+    if meandiff != 0:
```
jsolbrig, line 275: I think that, given when this function will be called, we can remove these if statements. The only thing they end up doing is cause the function to error if we call it with identical objects for `object1` and `object2`. When removed, the function works as expected, even when the objects are identical.

geoips/plugins/modules/output_checkers/netcdf.py
biosafetylvl5: See issue #469

geoips/plugins/yaml/sectors/static/denver.yaml
```diff
@@ -20,9 +20,9 @@ spec:
...
   resolution:
-    - 100
-    - 100
+    - 500
+    - 500
   shape:
-    height: 2000
-    width: 2000
+    height: 400
+    width: 400
```
mindyls, line 27: Are we ok with this being our sample Denver sector?

mindyls (approved):
Assuming someone else tests this, I am good to go with it.

I did not test it locally.

geoips/interfaces/module_based/output_checkers.py
```diff
@@ -517,8 +517,11 @@ def gunzip_product(fname, is_comparison_product=False, clobber=False):
     # Otherwise, see if the .gz version exists.
     elif glob(fname + ".gz"):
         gz_fname = fname + ".gz"
-    LOG.info("**** Gunzipping product for comparisons")
-    LOG.info("gunzip %s", gz_fname)
+
+    message = ["Gunzipping product for comparisons"]
+    message.append(f"gunzip {gz_fname}")
+    log_with_emphasis(LOG.info, message)
```
evrose54, line 523: This won't work currently. You'll need to edit this line to ``log_with_emphasis(LOG.info, *message)`` or update log_with_emphasis to unpack a list if provided.

biosafetylvl5: Ah 😢 thanks @evrose54 - I keep making that mistake. Thanks for catching it!

biosafetylvl5: Fixed!

geoips/plugins/modules/output_checkers/netcdf.py
biosafetylvl5: Fixed!

evrose54 (changes requested):
See my other comment for the changes needed. Big thing is that python strings have no ``append`` function, so ``print_area_def`` will not work. It's not used anywhere which explains why the tests pass, but this wouldn't work nonetheless. Since it's not used anywhere, I suggest we just remove ``print_area_def`` from the codebase.

geoips/plugins/modules/procflows/single_source.py
```diff
@@ -361,11 +362,10 @@ def process_xarray_dict_to_output_format(
 
 def print_area_def(area_def, print_str):
     """Print area def."""
-    emphasis_str = "*" * 84
-    LOG.info(f"\n\n{emphasis_str}" f"\n***{print_str}\n{area_def}")
+    message = f"{print_str}\n{area_def}"
     for key, value in area_def.sector_info.items():
-        LOG.info(f"{key}: {value}")
-    LOG.info(emphasis_str)
+        message.append(f"{key}: {value}")
```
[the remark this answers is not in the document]
biosafetylvl5: I fixed this, but see @evrose54 's comment on how this function is never used below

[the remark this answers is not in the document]
jsolbrig: I removed the function. Until I saw what it did, I thought we should just move it elsewhere, but it doesn't do much and can just be rewritten if we want a similar function in the future.

geoips/plugins/yaml/sectors/static/denver.yaml
jsolbrig: Yeah, this is a little weird. A 2000x2000 km image for a city?

I just updated to be 250x250m resolution and 200x200 shape.

jsolbrig (approved):

evrose54 (approved):
Looks good to me. Thanks for making the appropriate updates that I requested!

Issue #455, opened 2024-03-19
CLI Stage Two


evrose54:
# Requested Update

### Description
With the new PR #444, which creates the initial framework and functionality for the GeoIPS CLI, we are in a state to continue to develop the CLI to perform more complex processes. Currently, the CLI supports ``geoips [config, get, list, run, test, validate]`` commands, however we'd like to modify these to expand their uses cases. This issue wants to refactor ``geoips run`` and ``geoips list`` to support a greater swath of actions that they'll be able to perform.

### Background and Motivation
We envision the ``geoips run`` command to actively call GeoIPS' ``run_procflow`` command, which executes process workflows with the provided arguments. Currently, ``geoips run`` is called via ``geoips run <script_name> <-p> <package_name>``. While this works, it's not how we want it to proceed for the foreseeable future. This issue requests that we modify the functionality of ``geoips run`` to actively call ``run_procflow`` instead.

We don't want to replace the functionality of ``geoips list``, rather we want to expand on it to allow users to select what type of output they want sent to the terminal. Depending on the subcommand stemming from ``geoips list``, you'll get a generated list of data with static column headers. We should modify the ``geoips list`` command to support additional flags such as ``--column [header1, header2, ...]`` so that users can specify exactly what they'd like outputted. We'd also like to support flags such as ``--short`` or ``--long``, which will correspond to header lists that will be outputted appropriately.

### Code to demonstrate issue
*  See #444, and the information included in that PR for examples of how to execute the GeoIPS CLI

### Checklist for Completion
* [ ] Refactor ``geoips run`` to actively call ``run_procflow`` instead.
* [ ] Refactor ``geoips list`` to support ``--column [header1, header2, ...]``
* [ ] Refactor ``geoips list`` to support ``--short``
* [ ] Refactor ``geoips list`` to support ``--long``

evrose54:
For ``geoips run`` we additionally need to consider arguments that are used by other procflow plugins such as data_fusion and config_based. While single_source and config_based consider the same arguments, this might not always be the case. ``geoips.commandline.run_procflow.py:main``'s signature allows for a specific ``get_commandline_args_func`` to be specified. If this function is specified, we will add arguments to an ``argparse.ArgumentParser`` object which will take on arguments specific to that current procflow. Considering we don't know what procflow is being ran until we've parsed the ArgumentParser's Namespace, we have a few options.

1:
[...]

Issue #37, opened 2022-08-11
Create a useful CLI for GeoIPS


jsolbrig:
# Requested Update

### Description
Currently, doesn't really have a CLI to speak of. I would like to unify all GeoIPS commands under a single `geoips` command with subcommands. Additional functionality should also be exposed this way.

In addition, we should examine the command line arguments to determine if they can be cleaned up. There are a huge number of them, especially for `run_procflow`.

Lastly, we should work to update the descriptions of the available commands. `run_procflow` is simply described as "Run data file processing" which isn't very descriptive.

#### Current Commands
The current GeoIPS CLI is broken up across six commands. Those commands are not discoverable, they must simply be known from documentation. The current commands are:
- `run_procflow=geoips.commandline.run_procflow:main`: Run data file processing
- `convert_trackfile_to_yaml=geoips.commandline.convert_trackfile_to_yaml:main`: *Currently `-h` returns a ModuleNotFoundError.*
- `update_tc_tracks_database=geoips.commandline.update_tc_tracks_database:main`: *Currently `-h` actually runs the command rather than returning help info*
- `xml_to_yaml_sector=geoips.commandline.test_interfaces:main`: *Currently `-h` returns a ModuleNotFoundError.*
- `test_interfaces=geoips.commandline.test_interfaces:main`: *Currently `-h` actually runs the command rather than returning help info*
- `list_available_modules=geoips.commandline.list_available_modules:main`: *Currently `-h` actually runs the command rather than returning help info*

#### Suggested CLI
I propose that all commands fall under a main `geoips` command so that they can be discovered via `geoips -h/--help`. Below are a list of proposed commands.

- `run`: Works like `run_procflow`
- `list`: List a variety of different things based on the provided subcommand. One per interface module, probably.
  - `algorithms`: Return a list of available algorithms. Options could include `--by_type`, `--by_sensor`, `--sensor`, `--sensor_class`
  - `area_def_generators`: Return a list of area definition generators. (Maybe shorten to `ad_generators`?)
  - `coverage_checks`: Return a list of available coverage checkers.
  - `filename_formats`: Return a list of file name formats. (Maybe rename to `filename_formatters`?)
  - `interpolation`: Return a list of the interpolation methods. (Maybe rename to `interpolators`?)
  - `output_formats`: Return a list of available output formats. (Split up by type? Imagery, Data, etc?)
  - `procflows`: Return a list of available procflows. (Maybe these could be subcommands of `run`?)
[...]

PR #465, opened 2024-04-02
455 cli stage two

evrose54:
# Reviewer Checklist

# Related Issues
fixes NRLMMD-GEOIPS/geoips#37
fixes NRLMMD-GEOIPS/geoips#455
fixes NRLMMD-GEOIPS/geoips#571
fixes NRLMMD-GEOIPS/geoips#572
fixes NRLMMD-GEOIPS/geoips#578
fixes NRLMMD-GEOIPS/geoips#584
fixes NRLMMD-GEOIPS/geoips#585
fixes NRLMMD-GEOIPS/geoips#587
fixes NRLMMD-GEOIPS/geoips#596

partially addresses NRLMMD-GEOIPS/geoips#579
See https://github.com/NRLMMD-GEOIPS/data_fusion/pull/31

# Testing Instructions
To ensure all CLI functionality is running correctly, run ``pytest -v tests/unit_tests/commandline``. All tests should pass.

# Summary
With the addition of the GeoIPS CLI (https://github.com/NRLMMD-GEOIPS/geoips/pull/444) we now have access to
a wide array of commands to interact with GeoIPS. The purpose of this PR is to enhance
the GeoIPS CLI with new and improved functionality. Specifically this PR addresses the
``geoips run`` command and adds additional arguments (``--columns, --long``) to
``geoips list`` commands.

``geoips run`` now invokes the same process as what ``run_procflow`` and
``data_fusion_procflow`` do. We've made this change to allow process workflow
functionality to become part of the CLI rather than a separate commandline entrypoint
that was hard to discover. Eventually, it would be nice to move all independent
commandline entrypoints onto the CLI. Such as ``create_plugin_registries``,
``create_sector_image``, etc. To use ``geoips run``, simply replace ``run_procflow``
or ``data_fusion_procflow`` with ``geoips run <specified_procflow>``, keeping the
remaining arguments the exact same. We have deprecated the use of the old procflow calls
and the ``--procflow <specified_procflow>`` flag as that information is now obtained by
the ``geoips run <specified_procflow>`` call. Below is an example of what a new
procflow script looks like:

    geoips run single_source $GEOIPS_TESTDATA_DIR/test_data_noaa_aws/data/goes16/20200918/1950/* \
        --reader_name abi_netcdf \
        --product_name Infrared \
        --compare_path "$GEOIPS_PACKAGES_DIR/geoips/tests/outputs/abi.static.<product>.imagery_annotated" \
        --output_formatter imagery_annotated \
        --filename_formatter geoips_fname \
        --resampled_read \
        --logging_level info \
        --sector_list goes_east

As stated above, we also added new optional arguments to all ``geoips list`` commands.
These arguments are shared and are only specified once via the ``GeoipsListCommon``
class. The two new optional arguments are ``--long`` and ``--columns <col_header> ...``.
``--long`` is the default for all ``geoips list`` commands and specifies that all
headers should be outputted. ``--columns <col_header> ...`` Specifies which columns
you'd like outputted via the ``geoips list`` command. For example, if we ran
``geoips list interface algorithms``, we'd get the ``--long`` format outputted. Running
``geoips list interface algorithms --columns package interface plugin_type relpath``
provides us table output with only those specified columns. Available columns differ
by list command.

*Files Modified*

    added: tests/integration_tests/cli_dummy_integration.sh
    added: tests/scripts/cli_dummy_script.sh
    modified: geoips/commandline/ancillary_info/cmd_instructions.json
    modified: geoips/commandline/ancillary_info/cmd_instructions.yaml
    modified: geoips/commandline/args.py
    modified: geoips/commandline/commandline_interface.py
    modified: geoips/commandline/geoips_list.py
    modified: geoips/commandline/geoips_run.py
    modified: geoips/commandline/run_procflow.py
    modified: geoips/create_plugin_registries.py
    modified: geoips/geoips_utils.py
    modified: pyproject.toml
    modified: tests/scripts/abi.config_based_exclude_goes17.sh
    modified: tests/scripts/abi.config_based_output.sh
    modified: tests/scripts/abi.config_based_output_low_memory.sh
    modified: tests/scripts/abi.static.Infrared.imagery_annotated.sh
    modified: tests/scripts/abi.static.Visible.imagery_annotated.sh
    modified: tests/scripts/ahi.tc.WV.geotiff.sh
    modified: tests/scripts/amsr2.config_based_no_compare.sh
    modified: tests/scripts/amsr2.config_based_no_compare_full.sh
    modified: tests/scripts/amsr2.config_based_overlay_output.sh
    modified: tests/scripts/amsr2.config_based_overlay_output_low_memory.sh
    modified: tests/scripts/amsr2.tc.89H-Physical.imagery_annotated.sh
    modified: tests/scripts/amsr2_ocean.tc.windspeed.imagery_clean.sh
    modified: tests/scripts/amsub_mirs.tc.183-3H.imagery_annotated.sh
    modified: tests/scripts/ascat_knmi.tc.windbarbs.imagery_windbarbs_clean.sh
    modified: tests/scripts/ascat_low_knmi.tc.windbarbs.imagery_windbarbs.sh
    modified: tests/scripts/ascat_noaa_25km.tc.windbarbs.imagery_windbarbs.sh
    modified: tests/scripts/ascat_noaa_50km.tc.wind-ambiguities.imagery_windbarbs.sh
    modified: tests/scripts/ascat_uhr.tc.nrcs.imagery_clean.sh
    modified: tests/scripts/ascat_uhr.tc.wind-ambiguities.imagery_windbarbs.sh
    modified: tests/scripts/ascat_uhr.tc.windbarbs.imagery_windbarbs.sh
    modified: tests/scripts/ascat_uhr.tc.windspeed.imagery_clean.sh
    modified: tests/scripts/atms.tc.165H.netcdf_geoips.sh
    modified: tests/scripts/documentation_imagery.sh
    modified: tests/scripts/ewsg.static.Infrared.imagery_clean.sh
    modified: tests/scripts/gmi.tc.89pct.imagery_clean.sh
    modified: tests/scripts/hy2.tc.windspeed.imagery_annotated.sh
    modified: tests/scripts/imerg.tc.Rain.imagery_clean.sh
    modified: tests/scripts/mimic_coarse.static.TPW-CIMSS.imagery_annotated.sh
    modified: tests/scripts/mimic_fine.tc.TPW-PWAT.imagery_annotated.sh
    modified: tests/scripts/modis.Infrared.unprojected_image.sh
    modified: tests/scripts/oscat_knmi.tc.windbarbs.imagery_windbarbs.sh
    modified: tests/scripts/saphir.tc.183-3HNearest.imagery_annotated.sh
    modified: tests/scripts/sar.tc.nrcs.imagery_annotated.sh
    modified: tests/scripts/seviri.WV-Upper.unprojected_image.sh
    modified: tests/scripts/smap.tc.windspeed.imagery_clean.sh
    modified: tests/scripts/smap.unsectored.text_winds.sh
    modified: tests/scripts/smos.tc.sectored.text_winds.sh
    modified: tests/scripts/ssmi.tc.37pct.imagery_clean.sh
    modified: tests/scripts/ssmis.color91.unprojected_image.sh
    modified: tests/scripts/viirsclearnight.Night-Vis-IR-GeoIPS1.unprojected_image.sh
    modified: tests/scripts/viirsday.tc.Night-Vis-IR.imagery_annotated.sh
    modified: tests/scripts/viirsmoon.tc.Night-Vis-GeoIPS1.imagery_clean.sh
    modified: tests/unit_tests/commandline/test_geoips_list_interface.py
    modified: tests/unit_tests/commandline/test_geoips_list_packages.py
    modified: tests/unit_tests/commandline/test_geoips_list_scripts.py
    modified: tests/unit_tests/commandline/test_geoips_list_test_datasets.py
    modified: tests/unit_tests/commandline/test_geoips_list_unit_tests.py
    modified: tests/unit_tests/commandline/test_geoips_run.py
    modified: tests/unit_tests/commandline/test_geoips_test_script.py

# Output
Normal ``geoips list interface algorithms``
```
(geoips) (base) [evan@spring geoips]$ geoips list interface algorithms
----------
algorithms
----------
╭───────────────────────┬──────────────────┬──────────────────┬───────────────────────┬──────────────────────┬────────────────┬────────────────────────────────╮
│ GeoIPS Package        │ Interface Name   │ Interface Type   │ Family                │ Plugin Name          │ Source Names   │ Relative Path                  │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ data_fusion           │ algorithms       │ module_based     │ xarray_dict_to_xarray │ stitched             │ N/A            │ plugins/modules/algorithms/sti │
│                       │                  │                  │                       │                      │                │ tched.py                       │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ my_package            │ algorithms       │ module_based     │ xarray_to_xarray      │ pmw_89test           │ N/A            │ plugins/modules/algorithms/pmw │
│                       │                  │                  │                       │                      │                │ _89test.py                     │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips_glm            │ algorithms       │ module_based     │ xarray_dict_to_xarray │ ged                  │ N/A            │ plugins/modules/algorithms/ged │
│                       │                  │                  │                       │                      │                │ .py                            │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips_glm            │ algorithms       │ module_based     │ xarray_dict_to_xarray │ glm_alg              │ N/A            │ plugins/modules/algorithms/glm │
│                       │                  │                  │                       │                      │                │ _alg.py                        │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips_glm            │ algorithms       │ module_based     │ xarray_to_xarray      │ gremlinize_goes_kyle │ N/A            │ plugins/modules/algorithms/gre │
│                       │                  │                  │                       │                      │                │ mlinize_goes_kyle.py           │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips_glm            │ algorithms       │ module_based     │ xarray_to_xarray      │ gremlinize_goes      │ N/A            │ plugins/modules/algorithms/gre │
│                       │                  │                  │                       │                      │                │ mlinize_goes.py                │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips_plugin_example │ algorithms       │ module_based     │ list_numpy_to_numpy   │ single_channel_test  │ N/A            │ plugins/modules/algorithms/sin │
│                       │                  │                  │                       │                      │                │ gle_channel_test.py            │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ list_numpy_to_numpy   │ single_channel       │ N/A            │ plugins/modules/algorithms/sin │
│                       │                  │                  │                       │                      │                │ gle_channel.py                 │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ list_numpy_to_numpy   │ pmw_color89          │ N/A            │ plugins/modules/algorithms/pmw │
│                       │                  │                  │                       │                      │                │ _tb/pmw_color89.py             │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ list_numpy_to_numpy   │ pmw_37pct            │ N/A            │ plugins/modules/algorithms/pmw │
│                       │                  │                  │                       │                      │                │ _tb/pmw_37pct.py               │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ list_numpy_to_numpy   │ pmw_color37          │ N/A            │ plugins/modules/algorithms/pmw │
│                       │                  │                  │                       │                      │                │ _tb/pmw_color37.py             │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ list_numpy_to_numpy   │ pmw_89pct            │ N/A            │ plugins/modules/algorithms/pmw │
│                       │                  │                  │                       │                      │                │ _tb/pmw_89pct.py               │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ list_numpy_to_numpy   │ windbarbs            │ N/A            │ plugins/modules/algorithms/sfc │
│                       │                  │                  │                       │                      │                │ _winds/windbarbs.py            │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ list_numpy_to_numpy   │ Night_Vis            │ N/A            │ plugins/modules/algorithms/vis │
│                       │                  │                  │                       │                      │                │ ir/Night_Vis.py                │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ list_numpy_to_numpy   │ Night_Vis_GeoIPS1    │ N/A            │ plugins/modules/algorithms/vis │
│                       │                  │                  │                       │                      │                │ ir/Night_Vis_GeoIPS1.py        │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ list_numpy_to_numpy   │ Night_Vis_IR         │ N/A            │ plugins/modules/algorithms/vis │
│                       │                  │                  │                       │                      │                │ ir/Night_Vis_IR.py             │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────┼──────────────────────┼────────────────┼────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ list_numpy_to_numpy   │ Night_Vis_IR_GeoIPS1 │ N/A            │ plugins/modules/algorithms/vis │
│                       │                  │                  │                       │                      │                │ ir/Night_Vis_IR_GeoIPS1.py     │
╰───────────────────────┴──────────────────┴──────────────────┴───────────────────────┴──────────────────────┴────────────────┴────────────────────────────────╯
```
Listing specified columns of the output above. ``geoips list interface algorithms --columns package interface plugin_type relpath``
```
(geoips) (base) [evan@spring geoips]$ geoips list interface algorithms --columns package interface plugin_type relpath
----------
algorithms
----------
╭───────────────────────┬──────────────────┬──────────────────┬───────────────────────────────────────────────────────╮
│ GeoIPS Package        │ Interface Name   │ Interface Type   │ Relative Path                                         │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ data_fusion           │ algorithms       │ module_based     │ plugins/modules/algorithms/stitched.py                │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ my_package            │ algorithms       │ module_based     │ plugins/modules/algorithms/pmw_89test.py              │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips_glm            │ algorithms       │ module_based     │ plugins/modules/algorithms/ged.py                     │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips_glm            │ algorithms       │ module_based     │ plugins/modules/algorithms/glm_alg.py                 │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips_glm            │ algorithms       │ module_based     │ plugins/modules/algorithms/gremlinize_goes_kyle.py    │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips_glm            │ algorithms       │ module_based     │ plugins/modules/algorithms/gremlinize_goes.py         │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips_plugin_example │ algorithms       │ module_based     │ plugins/modules/algorithms/single_channel_test.py     │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ plugins/modules/algorithms/single_channel.py          │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ plugins/modules/algorithms/pmw_tb/pmw_color89.py      │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ plugins/modules/algorithms/pmw_tb/pmw_37pct.py        │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ plugins/modules/algorithms/pmw_tb/pmw_color37.py      │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ plugins/modules/algorithms/pmw_tb/pmw_89pct.py        │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ plugins/modules/algorithms/sfc_winds/windbarbs.py     │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ plugins/modules/algorithms/visir/Night_Vis.py         │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ plugins/modules/algorithms/visir/Night_Vis_GeoIPS1.py │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ plugins/modules/algorithms/visir/Night_Vis_IR.py      │
├───────────────────────┼──────────────────┼──────────────────┼───────────────────────────────────────────────────────┤
│ geoips                │ algorithms       │ module_based     │ plugins/modules/algorithms/visir/Night_Vis_IR_GeoIPS1 │
│                       │                  │                  │ .py                                                   │
```


docs/source/releases/v1_12_2a0.rst
```diff
@@ -10,18 +10,19 @@
...
  | # # # https://github.com/U-S-NRL-Marine-Meteorology-Division/
 
-Version 1.12.2a0 (2024-02-21)
+Version 1.12.2a0 (2024-02-19)
 *****************************
 
+* Enhancements
 
+  * Initial Stage of the GeoIPS CLI
+  * CLI Stage Two - new implementation of 'geoips run' and more args for 'geiops list'
```
evrose54, line 19: Not ready for merge until #444 has been merged.

evrose54: But I will say that this PR is ready for review. All functionality has been added and all tests pass.

evrose54: This has been merged with the most recent updates to #444 and is ready for reviewal / merge process.

geoips/commandline/commandline_interface.py
```diff
@@ -0,0 +1,119 @@
...
+    -------
+    legacy: bool
+        - The truth value as to whether or not a legacy procflow call was used
+    """
+    defined_procflow = None
+    if "run_procflow" == basename(sys.argv[0]):
+        entrypoint = "run_procflow"
+        defined_procflow = "single_source"
+
+    elif "data_fusion_procflow" == basename(sys.argv[0]):
```
jsolbrig, line 90: I think it is conventional to put the variable on the left side of comparisons. This is nitpicky, but it tripped me up.

geoips/commandline/geoips_command.py
```diff
@@ -0,0 +1,495 @@
...
+                combined_name = self.subcommand_name
+                parent_parsers = []
+            else:
+                combined_name = f"{self.parent.subcommand_name}_{self.subcommand_name}"
+                # We need to create a Geoips<cmd>Common Class for arguments
+                # that are shared between common commands. For example, we've created
+                # a 'GeoipsListCommon' class which adds arguments that will be shared
+                # by each GeoipsList<sub-cmd> class. Ie. if GeoipsListCommon has
+                # arguments --package, --columns, etc., and all of those arguments
+                # would be inherited by each GeoipsList<sub-cmd>
```
jsolbrig, line 80: Is there an issue for this?

tests/unit_tests/commandline/cli_top_level_tester.py
```diff
@@ -76,6 +77,59 @@ def test_datasets(self):
...
+            - The caputured output of the terminal
+        headers: dict of strings
+            - The Corresponding Header, column_val dict to parse through
+        selected_columns: list of strings or None
+            - If None, asser that every Header key is in the output. Otherwise check
+              that the corresponding Header key, column_val is in the output
+        """
+        for header in headers:
+            if selected_cols is None or headers[header] in selected_cols:
+                assert header in output
```
jsolbrig, line 131: I'm fairly sure that this should be updated to be `assert header in output or "has no plugins" in output` as was done [here](https://github.com/NRLMMD-GEOIPS/geoips/commit/906941b6213bf2fbb30b37f9b3ef0909eabd1c78#diff-11dc43f616ea37ea36fc61fe2fba69b276110395fee9f6853391eab0b61bab29R78)

tests/unit_tests/commandline/test_geoips_list_interfaces.py
```diff
@@ -61,25 +91,29 @@ def check_output(self, args, output):
             assert "To use, type `geoips list interfaces`" in output
         else:
             # The args provided are valid, so test that the output is actually correct
+            selected_cols = self.retrieve_selected_columns_from_list_command(args)
             if "-p" in args:
                 # Certain package has been requested, assert this was used alongside
                 # the "-i" ("--implemented") flag
                 assert "-i" in args
```
jsolbrig, line 98: I think we may want to rethink these arguments if, when one is required they are both required. This can probably go along with #586 which discusses fixing the language used for "implemented". 

Maybe we need to rethink what `geoips list interfaces` does a little bit. It seems like it should just list the available interfaces. Maybe listing which interfaces a package has plugins for should actually be a separate command.

tests/unit_tests/commandline/cli_top_level_tester.py
```diff
@@ -76,6 +77,59 @@ def test_datasets(self):
             ]
         return self._test_datasets
 
+    def retrieve_selected_columns_from_list_command(self, args):
+        """If --columns was used for a 'list' command, retrieve the list of columns.
```
jsolbrig, line 81: Is this only used for "list" commands? If so, should this be moved to tests relating to the "list" command or do we anticipate having other CLI commands that will output columns? If so, should this be renamed to just be something like "retrieve_table_columns"?

geoips/commandline/args.py
```diff
@@ -512,14 +514,18 @@ def add_args(parser, arglist=None):
         title="Processing workflow specifications"
     )
     if arglist is None or "procflow" in arglist:
+        if legacy:
+            help_str = """Specify procflow that should be followed for this file,
+                            located in geoips*.plugins.modules.procflows.
+                            myprocflowname.myprocflowname.
+                            The procflow string should be the procflow module
+                            name (no .py)"""
```
jsolbrig, line 522: This formatting can cause problems sometimes since the resulting string will contain a bunch of extra spaces. I would prefer to use something like this:
```python
            help_str = (
                "Specify procflow that should be followed for this file, "
                "located in geoips*.plugins.modules.procflows. "
                "myprocflowname.myprocflowname. "
                "The procflow string should be the procflow module "
                "name (no .py)"
            )
```

jsolbrig: Are there other instances of this kind of string formatting? If there is only this one let's fix it here. If there are more, let's create an issue.

jsolbrig: Alternatively, should this go in the same YAML file as the command help information?

pyproject.toml
```diff
@@ -150,7 +150,7 @@ geoips = "geoips"
 convert_trackfile_to_yaml = "geoips.commandline.convert_trackfile_to_yaml:main"
 create_sector_image = "geoips.commandline.create_sector_image:main"
 list_available_plugins = "geoips.commandline.list_available_plugins:main"
-run_procflow = "geoips.commandline.run_procflow:main"
+run_procflow = "geoips.commandline.commandline_interface:main"
```
jsolbrig, line 153: Have we made the same update to `pyproject.toml` from `data_fusion`?

geoips/commandline/commandline_interface.py
```diff
@@ -83,9 +84,50 @@ def execute_command(self):
...
+
+    Returns
+    -------
+    legacy: bool
+        - The truth value as to whether or not a legacy procflow call was used
+    """
+    defined_procflow = None
+    if "run_procflow" == basename(sys.argv[0]):
+        entrypoint = "run_procflow"
+        defined_procflow = "single_source"
```
jsolbrig, line 104: Does this also handle `config_based`?

jsolbrig: Oh, never mind. I see how this is done below. This could be more clear code-wise. Maybe we can have a separate function that determines the procflow rather than setting it, then re-setting it?

pyproject.toml
evrose54: Yup! See NRLMMD-GEOIPS/data_fusion#31.

tests/unit_tests/commandline/cli_top_level_tester.py
evrose54: You're correct. This has since been fixed. Changed it to ``assert header in output or "has no" in output`` as there are a few list commands that will output ``has no <artifact>`` depending on the command called.

tests/unit_tests/commandline/test_geoips_list_interfaces.py
evrose54: I fixed this portion. There was a conflict with this PR vs #444, in which this portion persisted. I've just removed this portion, since if ``-p`` is provided it's implied it's ``implemented``. We can fix this in #586 once we decide exactly how we want to handle this.

tests/unit_tests/commandline/cli_top_level_tester.py
evrose54: I could foresee that this might be used by the ``get / describe`` command in the future. Currently it's only list commands, but if we wanted to get a plugin, only retrieving information about its ``package, family, and docstring``, then we could apply this to those commands as well.

geoips/commandline/args.py
evrose54: I think this is the only occurrence of this. Let me check to make sure. Since this is technically an argument supported by our legacy ``run_procflow`` calls, it's required to be here. We could duplicate this information in each ``GeoipsRun<procflow>`` command class, but that seems repetitive. Maybe we should just update the yaml instructions clarifying that you don't have to specify ``--procflow <procflow_name>`` in your test scripts if you use ``geoips run <procflow_name>``?

geoips/commandline/commandline_interface.py
evrose54: I can replace all occurrences of this with what you recommended above. Always learning new conventions! I'll let you know when all of these are changed.

geoips/commandline/commandline_interface.py
evrose54: The only reason it is implemented the way it is currently is that we don't currently enforce what procflows can be defined in the legacy ``--procflow <procflow_name>`` argument. By defaulting to either ``single_source`` or ``data_fusion``, we can ensure that a correct procflow is called, whether or not it works. @jsolbrig, we could check that if either ``run_procflow`` or ``data_fusion`` entry-points are in ``sys.argv``, and if ``--procflow`` is in ``sys.argv``, then we could enforce that it's one of ``config_based``, ``data_fusion``, or ``single_source``. Eventually we could just enforce ``order_based``.  What do you think?

geoips/commandline/args.py
evrose54: It might be a good idea to make an issue for this nonetheless @jsolbrig, as all arguments created in ``geoips.commandline.args:add_args`` are formatted in the ``"""text..."""`` format highlighted in the code section above.

geoips/commandline/commandline_interface.py
evrose54: Done!

geoips/commandline/commandline_interface.py
```diff
@@ -83,9 +84,50 @@ def execute_command(self):
...
+        sys.argv.insert(2, defined_procflow)
+        return True
+    return False
+
+
 def main():
     """Entry point for GeoIPS command line interface (CLI)."""
-    geoips_cli = GeoipsCLI()
+    legacy = support_legacy_procflows()
+    geoips_cli = GeoipsCLI(legacy)
```
jsolbrig, line 130: Wouldn't this need to be `GeoipsCLI(legacy=legacy)`? This looks like it would go to `instructions_dir`.

Maybe we need to add one integration test to ensure that legacy calls work correctly. Or a unit test to ensure that GeoipsCLI receives its arguments correctly.

geoips/commandline/commandline_interface.py
jsolbrig: I'm thinking about this. It just occurred to me that this might not work if data_fusion isn't installed. Maybe we could test for a valid procflow by comparing against the available procflows? Do we have access to the available plugins at this point in the code?

jsolbrig: Discussed on a call:

- Will update such that when `geoips run <procflow> --legacy` is called, this will check the specified procflow against the list of installed procflows when running in legacy mode. If the procflow doesn't exist, an error will be raised.
- Will also update such that when `geoips run` is called without the `--legacy` option a `NotImpelemented` error will be raised. The error will indicate that, eventually, `geoips run` will always call the "order-based" procflow, but the order-based procflow is currently under development.

evrose54: See #598.

geoips/commandline/commandline_interface.py
evrose54: Yup. Let me do this real quick. See #598 for a discussion about legacy procflows.

geoips/commandline/geoips_command.py
```diff
@@ -71,13 +71,32 @@ def __init__(self, parent=None):
...
+            # that are shared between common commands. For example, we've created
+            # a 'GeoipsListCommon' class which adds arguments that will be shared
+            # by each GeoipsList<sub-cmd> class. Ie. if GeoipsListCommon has
+            # arguments --package, --columns, etc., and all of those arguments
+            # would be inherited by each GeoipsList<sub-cmd>
+            if "list" in self.combined_name:
+                parent_parsers = [GeoipsListCommon().subcommand_parser]
+            else:
+                parent_parsers = []
+
```
jsolbrig, line 99: We should consider how we can use `parser.parse_known_args()` to achieve this. I'll create an issue before merging this PR.

geoips/commandline/geoips_command.py
```diff
@@ -303,59 +331,192 @@ def _get_registry_by_interface_and_package(self, interface, package_name):
...
 
-    def _print_plugins_short_format(self, interface, interface_registry):
-        """Print the plugins under a certain interface alongside ancillary information.
+    def _get_headers_by_command(self, args, default_headers):
+        """Retrieve the appropriate headers from a list command with the given args.
 
-        "Short Format" includes these pieces of information:
-            [GeoIPS Package, Interface, Plugin Family, Plugin Name, Relpath]
+        Headers are the top-row of the tabular output from 'list' commands. These are
+        customizable via optional args [--column, --short, --long] and we retrieve
```
jsolbrig, line 338: `--short` doesn't currently appear to be handled.

I think we've said that we need to discuss what is included in the default column set, `--short`, and `--long`. Do we have an issue for that?

geoips/commandline/geoips_command.py
```diff
@@ -303,59 +331,192 @@ def _get_registry_by_interface_and_package(self, interface, package_name):
...
+                if col not in list(default_headers.keys()):
+                    err_str = f"Column header '{col}', is not a valid header. Please "
+                    err_str += "select one or more of the following keys, which "
+                    err_str += "correspond to the appropriate value:\n"
+                    err_str += f"{default_headers}"
+                    self.subcommand_parser.error(err_str)
+                headers[col] = default_headers[col]
+        else:
+            # long has been set or was defaulted to; use the default headers
+            headers = default_headers
```
jsolbrig, line 365: So, long is the default? Should that always be the case? Should `default_headers` be called `available_columns` and have the superset of all available columns while `--long` provides a subset of those?

This is legitimately a question, not me trying to suggest this is how I'd like it to be.

geoips/commandline/geoips_command.py
```diff
@@ -303,59 +331,192 @@ def _get_registry_by_interface_and_package(self, interface, package_name):
...
+        mutex_group.add_argument(
+            "--columns",
+            "-c",
+            type=str,
+            nargs="+",
+            default=None,
+            help="""Specific Headers of Data you'd like to see listed.
+                    For more in formation on headers available, run
+                    'geoips list <sub-cmd> <positional_args> --columns help'.
+                    """,
```
jsolbrig, line 517: When I call `geoips list interfaces --columns help` I get the following error. I don't think that `help` is currently being handled in `_get_headers_by_command`.

```
usage: To use, type `geoips list interfaces`. Optionally includes `-i` and `-p` flags for
seeing what has been implemented (-i) in a certain package (-p), or all packages.
For example, to see what interfaces have been implemented in a certain package,
run `geoips list interfaces -i -p <pkg_name>`, or you can remove the `-p` flag to
see what's been implemented in every package.
To use, type `geoips list <sub-cmd> <sub-sub-cmd>`. interfaces: error: Column header 'help', is not a valid header. Please select one or more of the following keys, which correspond to the appropriate value:
{'package': 'GeoIPS Package', 'plugin_type': 'Interface Type', 'interface': 'Interface Name', 'supported_families': 'Supported Families', 'docstring': 'Docstring', 'abspath': 'Absolute Path'}
```

geoips/commandline/geoips_command.py
```diff
@@ -303,59 +331,192 @@ def _get_registry_by_interface_and_package(self, interface, package_name):
...
+        -------
+        table_data: 2D List of Strings
+            - The corresponding data we will output, where each sub-list is ordered
+              by headers.
+        """
+        table_data = []
+        if interface.name == "products":
+            for plugin_key in sorted(interface_registry.keys()):
+                product_dict = interface_registry[plugin_key]
+                table_data += self._get_entry(product_dict, headers)
```
jsolbrig, line 440: This is making me realize that we're not handling yaml plugins whose families are "list" correctly. Currently the only interface for which we implement lists of plugins is the Products interface, however, I always intended for the "list" family to be usable by any yaml-based plugin interface. So we could have lists of sectors, for instance.

This means that we should be handling plugins whose family is "list" more generally and they should be handled in the plugin registry rather than needing to be handled by the CLI. I'm not sure what the correct way to handle this is, but it's something that we need to work on. The difficulty with the Product plugins is that the same product name can exist for multiple source_names.

This is something that we should discuss and come up with a solution for.

geoips/commandline/geoips_command.py
```diff
@@ -303,59 +331,192 @@ def _get_registry_by_interface_and_package(self, interface, package_name):
...
+        else:
+            table_data += self._get_entry(interface_registry, headers)
+        return table_data
+
+    def _get_entry(self, plugin_dict, headers):
+        """Retrieve the appropriate plugin entry given a list of valid headers.
```
jsolbrig, line 446: Is this "retrieving"? It seems like it is "getting" or "appending".

It's not really clear from the docstring what this does or how/where it is used.

geoips/commandline/geoips_command.py
```diff
@@ -303,59 +331,192 @@ def _get_registry_by_interface_and_package(self, interface, package_name):
...
+            default=None,
+            help="""Specific Headers of Data you'd like to see listed.
+                    For more in formation on headers available, run
+                    'geoips list <sub-cmd> <positional_args> --columns help'.
+                    """,
+        )
+
+    def __call__(self, args):
+        """Exectutable function for GeoipsListCommon Class. Not implemented."""
+        pass
```
jsolbrig, line 522: Why is this an executable command?

geoips/commandline/geoips_list.py
```diff
@@ -275,10 +319,15 @@ def list_implemented_interfaces(self, package_name):
...
-            headers = ["GeoIPS Package", "Interface Type", "Interface Name"]
+                    interface_entry = []
+                    for header in list(headers.keys()):
+                        if header == "package":
+                            interface_entry.append(plugin_package_name)
+                        elif header == "plugin_type":
+                            interface_entry.append(interface_type)
+                        elif header == "interface":
+                            interface_entry.append(interface_name)
+                    interface_data.append(interface_entry)
```
jsolbrig, line 330: It seems like there is a lot of repetition of this kind of thing. Could this be centralized somewhere in a single function that can be called with the requested columns?

geoips/commandline/geoips_run.py
```diff
@@ -3,36 +3,89 @@
...
+                "procflow is being executed via '--procflow <procflow_name>' or "
+                "use the supported procflow call 'geoips run <procflow_name>'"
+            )
+            self.subcommand_parser.error(err_str)
+        elif args.procflow is None:
+            # If None, set to 'data_fusion'. We don't want users to have to specify
+            # what procflow will be used as it is specified in
+            # 'geoips run data_fusion'.
+            args.procflow = "data_fusion"
+        main(ARGS=args)
```
jsolbrig, line 77: Huh, that's confusing. I didn't realize that `data_fusion_procflow` actually called `main` from `run_procflow`. That seems pretty circuitous... Not worth fixing, just saying it seems weird.

geoips/commandline/geoips_command.py
evrose54: ``--short`` is not handled. There's a topic in #455 which discusses that we'd like to support it, but I didn't add it since we haven't discussed what we'd like included in its output. Personally, I think a lot of the ``list`` commands are already short in their output, and that the ``--columns`` flag is enough to get a filtered output if needed. Happy to implement that in this PR if we'd prefer though. 

All in all, I don't see the use for it, but if we all would prefer to have it, I'm happy to implement it.

geoips/commandline/geoips_command.py
evrose54: Yes, ``--long`` is the default. I'd assume that's what we'd always want to be the case. Give the user the most information about an artifact as possible, and if it's too much, they can filter using the ``--columns`` command. We can rename that to ``available_columns`` if you'd like.

geoips/commandline/geoips_command.py
evrose54: This will be fixed in #478 which includes update to how we access product plugins. Internally we'll search for them as a plugin name of ``source_name.product_name``, and access them via ``get_plugin`` by ``products.get_plugin(source_name, product_name)``. If we do this for every yaml that is of family ``list``, there shouldn't be a problem in the future.

geoips/commandline/geoips_command.py
evrose54: I can update this docstring.

geoips/commandline/geoips_command.py
evrose54: Executable commands have the ability to add arguments. Since we want shared arguments between all of the list commands, this was the only way to do this via our Abstract Class and argparse's implementation of sharing arguments.

geoips/commandline/geoips_run.py
evrose54: Agreed that it's weird. I just followed the way it was implemented currently, but I agree that it seems like a loop. Hopefully order_based will fix this.

geoips/commandline/geoips_list.py
evrose54: This will likely need a lot of refactoring and a lot of file changes. I'll make an issue for this one.

evrose54: See #602.

geoips/commandline/geoips_command.py
evrose54: This might be a good idea to tackle this in issue #580

geoips/commandline/commandline_interface.py
jsolbrig: Thanks, @evrose54, looks good!

geoips/commandline/geoips_command.py
jsolbrig: Can you create an issue to just discuss how to implement options for different scopes? I think we can make use of `ArgumentParser.parse_known_args` to help implement this.

Thinking about it, maybe there is already an issue?

tests/unit_tests/commandline/cli_top_level_tester.py
jsolbrig: If we foresee using this for other commands, let's rename this to just `retrieve_selected_columns` and update the docstring to remove references to the `list` command. 

If we only foresee using this for `list`, let's move it into the tests relating to `list`.

tests/unit_tests/commandline/test_geoips_list_interfaces.py
jsolbrig: Sounds good. Thanks!

geoips/commandline/args.py
jsolbrig: Okay, can you please create that issue and resolve this comment when done?

geoips/commandline/geoips_command.py
jsolbrig: Unless it already exists, can you create a new issue discussing this and reference #580 from it? They're separate but related issues I think. #580 probably depends on this issue.

In this issue, we want to develop a method of adding parser options at different scopes including global (i.e. on the top-level `geoips` command) and for any command, sub-command, sub-sub-command, etc. I think we can probably achieve this somehow using `ArgumentParser.parse_known_args()`.

geoips/commandline/geoips_command.py
jsolbrig: I think I'll make which columns to display a discussion point for our visit to NRL. I do think I'd like to switch this to `available_columns` or `all_columns` or even just `columns` if it doesn't conflict with anything else.

geoips/commandline/geoips_command.py
jsolbrig: Okay, we should talk about what the appropriate functionality is here. Maybe we should have a separate flag for `--list-columns` rather than offering `--columns help`?

geoips/commandline/geoips_command.py
jsolbrig: Got it, thanks!

geoips/commandline/geoips_command.py
jsolbrig: Okay, this is fine for now. I think we'll probably discuss this more when we discuss the inheritance structure as @biosafetylvl5 has suggested in the past. I'll just resolve this for now since it will get handled in the future, one way or another.

geoips/commandline/geoips_command.py
jsolbrig: I say more about this here: https://github.com/NRLMMD-GEOIPS/geoips/pull/465#discussion_r1619204653

geoips/commandline/geoips_command.py
jsolbrig: Great, thanks!

geoips/commandline/geoips_command.py
```diff
@@ -303,59 +331,201 @@ def _get_registry_by_interface_and_package(self, interface, package_name):
...
+        table_data = []
+        if interface.name == "products":
+            for plugin_key in sorted(interface_registry.keys()):
+                product_dict = interface_registry[plugin_key]
+                table_data += self._get_entry(product_dict, headers)
+        else:
+            table_data += self._get_entry(interface_registry, headers)
+        return table_data
+
+    def _get_entry(self, plugin_dict, headers):
```
jsolbrig, line 446: Is this the correct name for this function? It looks like it returns the entire table rather than a single row so does "entry" make sense here?

geoips/commandline/geoips_command.py
jsolbrig: In that case, can you please remove `--short` form the help messages?

geoips/commandline/geoips_get.py
```diff
@@ -3,34 +3,35 @@
 Retrieves the appropriate family/interface/package/plugin based on the args provided.
 """
 
-from importlib import resources, import_module
+from importlib import metadata, resources, import_module
 
 import yaml
 
 from geoips.commandline.geoips_command import GeoipsCommand, GeoipsExecutableCommand
+from geoips.create_plugin_registries import format_docstring
```
jsolbrig, line 11: This seems like an odd location to be importing from in this module. I think a refactor is probably needed here. We can create an issue for it, but I think we need to examine why this is required and fix the root of the problem.

geoips/create_plugin_registries.py
```diff
@@ -45,6 +46,38 @@
...
+        # Regex pattern for subbing out "\n" but not ".\n"
+        pattern = r"(?<!\.)\n"
+        docstring = re.sub(
+            pattern,
+            " ",
+            docstring.strip().replace("\n\n", "\n"),
+        )
+    else:
+        docstring = docstring.strip().replace("\n\n", "\n")
+    return docstring
```
jsolbrig, line 78: This is a nitpick, but this would be better written as:

```python
if docstring:
    if use_regex:
        # Regex pattern for subbing out "\n" but not ".\n"
        pattern = r"(?<!\.)\n"
        docstring = re.sub(
            pattern,
            " ",
            docstring.strip().replace("\n\n", "\n"),
        )
    else:
        docstring = docstring.strip().replace("\n\n", "\n")
return docstring
```

jsolbrig: Another question. Much of the time we add multiple new lines purposefully, to improve readability. Why remove them here?

geoips/geoips_utils.py
```diff
@@ -391,3 +392,49 @@ def merge_nested_dicts(dest, src, in_place=True, replace=False):
...
+        json_dist = json.loads(dist_info)
+        if (
+            "dir_info" in json_dist.keys()
+            and "editable" in json_dist["dir_info"].keys()
+            and json_dist["dir_info"]["editable"]
+        ):
+            # If the 'editable' key exists and is True
+            return True
+    # Package is installed in non-editable mode
+    return False
```
jsolbrig, line 440: I don't think we need a unit test for this, but have you tested this with packages installed in both ways?

tests/integration_tests/cli_dummy_integration.sh
```diff
@@ -0,0 +1,15 @@
...
+# # # This program is free software: you can redistribute it and/or modify it under
+# # # the terms of the NRLMMD License included with this program. This program is
+# # # distributed WITHOUT ANY WARRANTY; without even the implied warranty of
+# # # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the included license
+# # # for more details. If you did not receive the license, for more information see:
+# # # https://github.com/U-S-NRL-Marine-Meteorology-Division/
+
+#!/bin/bash
+
+echo "I ran a dummy integration test."
```
jsolbrig, line 15: Is this intentionally checked in?

tests/unit_tests/commandline/cli_top_level_tester.py
```diff
@@ -125,33 +262,139 @@ def check_output(self, args, output):
...
+            # except RuntimeError as re:
+            #     print(re, file=sys.stderr)
+
+        # Retrieve the output printed from 'func'
+        stdout = stdout_capture.getvalue()
+        stderr = stderr_capture.getvalue()
+
+        # Close the StringIO Objects
+        stdout_capture.close()
+        stderr_capture.close()
```
jsolbrig, line 307: This looks like it should work, but is it duplicating `pytest.capsys` and other similar fixtures from pytest? https://docs.pytest.org/en/7.1.x/how-to/capture-stdout-stderr.html

No need to change this now but, if this does duplicate capabilities from pytest, create an issue to switch to the pytest fixtures.

tests/unit_tests/commandline/cli_top_level_tester.py
```diff
@@ -125,33 +262,139 @@ def check_output(self, args, output):
...
+                return False
+            case _ if any(["non_existent" in arg for arg in args]):
+                # Can't capture argparse.ArgumentError output using monkeypatch... yet
+                return False
+            case _ if ("--long" in args and "--columns" in args):
+                # Can't capture argparse.ArgumentError output using monkeypatch... yet
+                return False
+            case _:
+                # Monkeypatch works for the provided arguments!
+                return True
```
jsolbrig, line 357: I think that a lot of this could be fixed by using pytest's built-in methods of capturing stdout/stderr. See here: https://docs.pytest.org/en/7.1.x/how-to/capture-stdout-stderr.html

tests/unit_tests/commandline/test_geoips_list_interfaces.py
```diff
@@ -9,24 +9,48 @@
...
+                + [
+                    "-i",
+                    "-p",
+                    "geoips",
+                    "--columns",
+                    "interface",
+                    "plugin_type",
+                ]
+            )
             # Add argument list with an invalid command call ("-p" w/out "-i")
```
jsolbrig, line 48: Is this comment still correct?

tests/unit_tests/commandline/test_geoips_list_interfaces.py
```diff
@@ -61,32 +91,36 @@ def check_output(self, args, output):
             assert "To use, type `geoips list interfaces`" in output
         else:
             # The args provided are valid, so test that the output is actually correct
+            selected_cols = self.retrieve_selected_columns_from_list_command(args)
             if "-i" in args or "-p" in args:
```
jsolbrig, line 95: Can this still be either "-i" or "-p"?

geoips/commandline/geoips_run.py
```diff
@@ -3,36 +3,89 @@
...
 """
 
-from glob import glob
-from importlib import resources
-from os.path import basename
-from subprocess import call
+from geoips.commandline.args import add_args
+from geoips.commandline.run_procflow import main
+from geoips.commandline.geoips_command import GeoipsCommand, GeoipsExecutableCommand
+from data_fusion.commandline.args import add_args as data_fusion_add_args
```
jsolbrig, line 9: This assumes that data_fusion is installed and raises an error if it is not.

```python
> geoips -h
  File "/home/jsolbrig/anaconda3/envs/geoips-github/bin/geoips", line 5, in <module>
    from geoips.commandline.commandline_interface import main
  File "/local/home/jsolbrig/geoips/github/geoips/geoips/commandline/commandline_interface.py", line 19, in <module>
    from geoips.commandline.geoips_run import GeoipsRun
  File "/local/home/jsolbrig/geoips/github/geoips/geoips/commandline/geoips_run.py", line 9, in <module>
    from data_fusion.commandline.args import add_args as data_fusion_add_args
ModuleNotFoundError: No module named 'data_fusion'
```

jsolbrig: This is making me wonder whether the CLI should, itself, accept plugins. I think that after we finish the order-based procflow, though, it won't be necessary.

geoips/geoips_utils.py
evrose54: Yup! I tested with a combination of editable and non-editable, all editable, and none editable.

tests/integration_tests/cli_dummy_integration.sh
evrose54: These are used so we don't actually run a full integration_test. Just to make sure that the commands discover the appropriate test and can run them.

tests/scripts/cli_dummy_script.sh
```diff
@@ -0,0 +1,15 @@
...
+# # # This program is free software: you can redistribute it and/or modify it under
+# # # the terms of the NRLMMD License included with this program. This program is
+# # # distributed WITHOUT ANY WARRANTY; without even the implied warranty of
+# # # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the included license
+# # # for more details. If you did not receive the license, for more information see:
+# # # https://github.com/U-S-NRL-Marine-Meteorology-Division/
+
+#!/bin/bash
+
+echo "I ran a dummy test script."
```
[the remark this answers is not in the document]
evrose54: These are used so we don't actually run a full procflow. Just to make sure that the commands discover the appropriate test and can run them.

tests/unit_tests/commandline/cli_top_level_tester.py
```diff
@@ -76,13 +85,138 @@ def test_datasets(self):
...
+    def command_combinations(self):
+        """A 2d List of command combinations for a CLI command call.
 
-        Ie. if we were testing 'geoips list interface', this property would be every
-        possible combination of strings used to call 'geoips list interface'.
+        This can cover every possible combination to call a command or it can be a
+        subset of commands if they take a while to run.
+
+        For a shorter command such as 'geoips list interface', this property would be
+        every possible combination of strings used to call 'geoips list interface'.
```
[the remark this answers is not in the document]
evrose54: This is a good idea. I agree that we should create a separate issue for this; I'll try to write that up by COB or get that created tomorrow.

tests/unit_tests/commandline/test_geoips_list_interfaces.py
evrose54: No it's not. Nice catch and I can remove that.

tests/unit_tests/commandline/test_geoips_list_interfaces.py
evrose54: Yup!

tests/unit_tests/commandline/cli_top_level_tester.py
evrose54: I'll play around with this and see if ``pytest.capsys`` works in another PR. I'll create an issue for this as well.

tests/unit_tests/commandline/cli_top_level_tester.py
evrose54: I'll write an issue to explore ``pytest.capsys`` in another PR and create an issue for this.

geoips/create_plugin_registries.py
evrose54: This is the format that I found to be most readable when looking at the output. Otherwise we are stuck with the newline problem in the docstring attribute that was poorly formatted. We can discuss exactly what we want at a future point. Happy to change this to what we all agree with.

geoips/commandline/geoips_get.py
evrose54: I'll create an issue for this. I agree that we should talk about this more. This function was addressing #572.

geoips/commandline/geoips_command.py
evrose54: I think we should leave it as ``get_entry`` or change it to ``get_plugin_entry``. This would only be called ``get_plugin_entries`` for product plugins currently, but would change to ``get_entry`` / ``get_plugin_entry`` once the changes in #474 come into effect.

evrose54: Since they would not be ``abi --> product_plugins``, and would be changed to `single `abi.Infrared``entries and so on.

tests/unit_tests/commandline/cli_top_level_tester.py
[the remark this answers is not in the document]
evrose54: See #627.

tests/unit_tests/commandline/cli_top_level_tester.py
evrose54: See #628.

tests/unit_tests/commandline/cli_top_level_tester.py
evrose54: See #628.

geoips/commandline/geoips_run.py
evrose54: What would you like me to do about this one? Do we want to do an optional import, move this to data_fusion, or leave it as is?

tests/unit_tests/commandline/test_geoips_list_interfaces.py
evrose54: This has been removed BTW.

geoips/commandline/geoips_get.py
evrose54: See #630.

geoips/commandline/geoips_run.py
jsolbrig: I think that, for now, let's make this an optional import. What I'd like to do in the long-run, though, is add the ability to have CLI plugins.

So, for now, would you please change this to be an optional import and add an issue to discuss whether CLI plugins are needed. Maybe they'd be useless after creating the order-based procflow, though, so just a discussion issue, not an actual "do this" issue.

evrose54: Sure! I can do that. I'll create an issue discussing your comment above as well.

geoips/create_plugin_registries.py
evrose54: I also modified that function to what you posted above.

geoips/commandline/geoips_run.py
evrose54: See #631.

docs/source/userguide/command_line.rst
```diff
@@ -261,21 +290,40 @@ For an example of how to run this command, see below.
 
     geoips list test-datasets
 
+``list unit-tests`` is a list sub-command which retrieves a listing of unit tests from
+all, or a certain GeoIPS Package. For this command to find your unit tets, you must
+place the unit tests under ``<package_name>/tests/unit_tests/``. These test scripts can
+then be ran using ``pytest -v /path/to/<package_name/tests/unit_tests/<unit_test_dir>``.
+This command can only be ran if the specified plugin package[s] are installed in
+*editable* mode.
```
mindyls, line 298: Does it have to be in editable mode? Or does the repo just have to be cloned (ie, could clone the repo, then install in site-packages ?)

It doesn't really matter, just asking out of curiosity / for my own understanding.

Just respond, and resolve this conversation.

docs/source/userguide/command_line.rst
```diff
@@ -306,29 +354,69 @@ Run Command
...
+        --sector_list goes_east
+    retval=$?
+
+    exit $retval
+
+New CLI-based Procflow (abi.static.Infrared.imagery_annotated.sh)
+
+.. code-block:: bash
+
+    geoips run single_source $GEOIPS_TESTDATA_DIR/test_data_noaa_aws/data/goes16/20200918/1950/* \
```
mindyls, line 401: I'm glad these are easy to update :-)

docs/source/userguide/command_line.rst
```diff
@@ -341,45 +429,45 @@ and where / why that is happening.
...
-
-Once those tests have been placed in the appropriate location, follow the command below.
-
-::
-
-    geoips test unit-test command_line
-    geoips test unit-test command_line -n test_geoips_list_packages.py
-    geoips test unit-test -p <package_name> <unit_test_dir_name>
-    geoips test unit-test -p <package_name> <unit_test_dir_name> -n <script_name>
+    geoips test script --implemented <script_name> (no '-p' as this is only supported for 'geoips' package)
```
mindyls, line 470: Is this supposed to be --integration?

geoips/commandline/ancillary_info/cmd_instructions.yaml
```diff
@@ -204,23 +215,49 @@ instructions:
       - Unit Test Name
   geoips_run:
     help_str: |
-      Run a geoips process workflow (procflow), given a GeoIPS package and a script
-      name. This will do the exact same thing as being in the directory of a certain
-      script, and running it via `./<script_name>`. To see what scripts are available,
-      you can run `geoips list-scripts <-p>`, where `-p` can be used alongside a GeoIPS
-      Package name if you want scripts from a single package.
+      Run a specified geoips process workflow (procflow) plugin. See
```
mindyls, line 218: I think this is technically a "processing workflow"?  I don't think it's worth updating, just a note.

To be fair, the only reference to what procflow means I found in docs is https://github.com/NRLMMD-GEOIPS/geoips/blob/main/docs/source/devguide/software_requirements_specification.rst buried in a sub bullet.

geoips/commandline/commandline_interface.py
```diff
@@ -83,10 +85,90 @@ def execute_command(self):
...
+        # functionality
+        if "--procflow" in sys.argv:
+            # If '--procflow' was found in the command line arguments, loop through the
+            # arguments and grab the correct procflow.
+            for idx, arg in enumerate(sys.argv):
+                if arg == "--procflow" and len(sys.argv) > idx + 1:
+                    defined_procflow = sys.argv[idx + 1]
+        sys.argv[0] = sys.argv[0].replace(entrypoint, "geoips")
+        sys.argv.insert(1, "run")
+        sys.argv.insert(2, defined_procflow)
```
mindyls, line 142: Ahh, so this is just converting the explicit data_fusion_procflow call to run geoips data_fusion ?  Works for me.

docs/source/userguide/command_line.rst
evrose54: Yes it is. Nice catch! I can fix that.

geoips/commandline/commandline_interface.py
evrose54: Yup! A little hacky but it works for the time being.

docs/source/userguide/command_line.rst
evrose54: It only works when it's in editable mode, since we are interrogating the /tests... directory. If it's installed in site-packages, this won't work since we don't include that directory. Hope this explains things!

docs/source/userguide/command_line.rst
evrose54: Me too!

docs/source/userguide/command_line.rst
evrose54: Fixed!

tests/scripts/cli_dummy_script.sh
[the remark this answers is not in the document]
mindyls: @evrose54 i think Jeremy wanted you to add a comment to the file itself

docs/source/releases/v1_13_0a0.rst
```diff
@@ -16,6 +16,7 @@ Version 1.13.0a0 (2024-04-24)
 * Enhancements
 
   * Initial Stage of the GeoIPS CLI
+  * CLI Stage Two - new implementation of 'geoips run' and more args for 'geiops list'
```
mindyls, line 19: I would update the title of this PR to reflect these actual changes,  rather than just CLI stage 2. 

Also,  were the bits about editable mode just added error checking?  Could we get the same results by just checking if the tests directory exists ? (Ie, I don't think it's specifically that the package is installed in editable mode,  just that the full repository has been cloned to disk ?)

mindyls (approved):
Once all conversations have been reviewed/ resolved,  this is good with me.  Thanks!

geoips/commandline/geoips_run.py
jsolbrig: Thanks @evrose54

tests/scripts/cli_dummy_script.sh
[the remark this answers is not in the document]
evrose54: Oh lol yes I can add a comment to this file.

tests/integration_tests/cli_dummy_integration.sh
evrose54: I'll add a comment to this file as well.

geoips/commandline/geoips_command.py
evrose54: Yup! I'll remove that from the help message.

geoips/commandline/geoips_command.py
evrose54: See #637.

geoips/commandline/geoips_command.py
evrose54: See #637.

geoips/commandline/geoips_command.py
evrose54: Removed!

geoips/commandline/args.py
evrose54: See #638.

tests/unit_tests/commandline/cli_top_level_tester.py
evrose54: Done!

docs/source/releases/v1_13_0a0.rst
jsolbrig: After reviewing, I think this is good as-is. One quibble I have is that the same logic is used in four different locations to raise an appropriate error when a command that requires editable mode is called. I think it's fine for now, but I think we should go through the tests at some point to trim them down and reduce code duplication.

geoips/commandline/geoips_command.py
jsolbrig: I'd like to create an issue to discuss and fix the handling of columns. For now, in the interest of getting this merged today, I'm going to close this issue and just send a link to this convesration to @evrose54 so that one of us can create an issue on Monday.

jsolbrig (approved):

Issue #468, opened 2024-04-02
log_with_emphasis wraps filenames poorly


biosafetylvl5:
Long string without spaces, like filenames, are wrapped by `log_with_emphasis`. This produces results like this:

```
02_232024    log_setup.py:41   INTERACTIVE: ********************************************************************************
02_232024    log_setup.py:46   INTERACTIVE: ** BAD GeoIPS NetCDF files do not match within tolerance                      **
02_232024    log_setup.py:46   INTERACTIVE: ** output_product: /home/jsolbrig/geoips/github/outdirs/preprocessed/algorith **
02_232024    log_setup.py:46   INTERACTIVE: ** ms/Infrared_latitude_longitude/abi/goes-                                   **
02_232024    log_setup.py:46   INTERACTIVE: ** 16/goes_east/20200918/20200918.195020.goes-                                **
02_232024    log_setup.py:46   INTERACTIVE: ** 16.Infrared_latitude_longitude.goes_east.nc                                **
02_232024    log_setup.py:46   INTERACTIVE: ** compare_product: /home/jsolbrig/geoips/github/geoips/tests/outputs/abi.sta **
02_232024    log_setup.py:46   INTERACTIVE: ** tic.Infrared.netcdf_geoips/20200918.195020.goes-                           **
02_232024    log_setup.py:46   INTERACTIVE: ** 16.Infrared_latitude_longitude.goes_east.nc                                **
02_232024    log_setup.py:47   INTERACTIVE: ********************************************************************************
```

which are Very difficult to read

PR #473, opened 2024-04-05
468 log with emphasis wraps filenames poorly

biosafetylvl5:
* [x] NO REQUIRED **existing tests** (explain why not required)
* [x] Required **unit tests** added and pass for new/modified functionality
* [x] NO REQUIRED **integration tests** (explain why not required)
* [x] NO REQUIRED **documentation** (explain why not required)
* [x] Required **release notes** added for new/modified functionality
* [x] NO REQUIRED **updates to other repos** (explain why not required)
# Related Issues
fixes NRLMMD-GEOIPS/geoips#468

# Testing Instructions
* No exhaustive testing required beyond existing and unit tests, as functionality hasn't changed that would impact these

# Summary
* Fixed an issue with log_with_emphasis() function that poorly wrapped file paths over 74 characters in length.
* Used existing wrap() function's functionality for breaking long filenames.
* Added/updated unit tests
* Updated release notes to reflect the fix.

### DIFF
--- a/tests/unit_tests/commandline/test_log_setup.py
+++ b/tests/unit_tests/commandline/test_log_setup.py
@@ -10,7 +10,6 @@ import pytest
 import os
 import random
 import string
-from textwrap import wrap
 from geoips import logging
 from glob import glob
 from geoips.commandline.log_setup import log_with_emphasis
@@ -23,7 +22,21 @@ def generate_random_string(length):
     return "".join(random.choices(string.ascii_letters, k=length))
 
 
-def insert_word_like_space_to_string(str):
+def insert_word_like_spaces_to_string(str):
+    """
+    Modify the input string by inserting spaces at random locations to make
+    "words" of length 2-8.
+
+    Parameters:
+    - str (str): The input string to be modified.
+
+    Returns:
+    - str: The modified string with spaces inserted at random locations.
+
+    Example:
+    >>> insert_word_like_spaces_to_string("HelloWorld")
+    'He lloW or ld'
+    """
     loc = random.randint(2, 3)
     while loc < len(str):
         str = str[:loc] + " " + str[loc + 1 :]
@@ -35,7 +48,9 @@ def generate_random_messages():
     """Generate a random amount of messages with random length."""
     num_messages = 20
     messages = [
-        insert_word_like_space_to_string(generate_random_string(random.randint(5, 110)))
+        insert_word_like_spaces_to_string(
+            generate_random_string(random.randint(5, 110))
+        )
         for _ in range(num_messages)
     ]
     return messages
@@ -45,7 +60,6 @@ def generate_random_messages():
 def test_log_with_emphasis(message, caplog):
     """Pytest function for testing the output of 'log_with_emphasis'."""
     caplog.set_level(logging.INFO)
-    max_message_len = min(74, len(message))
     log_with_emphasis(LOG.info, message)
     assert (  # top/bottom of box is formmated correctly
         "*" * 9  # three for boarders, and min of 5 for string length
--- a/tests/unit_tests/commandline/test_log_setup.py
+++ b/tests/unit_tests/commandline/test_log_setup.py
@@ -103,20 +103,19 @@ def test_log_with_emphasis(message, caplog, test_all_lines_same_length=True):
     """
     caplog.set_level(logging.INFO)
     log_with_emphasis(LOG.info, message)
-    assert (  # top/bottom of box is formmated correctly
-        "*" * 9 in caplog.text  # three for borders, and min of 5 for string length
+    assert (  # top of box is formmated correctly
+        "*" * 9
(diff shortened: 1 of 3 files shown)


evrose54 (changes requested):
I generally tend to never use keywords that are reserved in the python namespace. So, for ``geoips/tests/unit_tests/commandline/test_log_setup:insert_word_like_spaces_into_string``, I'd replace the ``str`` argument with ``string`` or ``message`` or something along those lines. Not sure if other people agree, but this is generally something I stick to strictly. Other than that, this PR looks good to me.

biosafetylvl5:
Fixed the `str` conflict! Thanks @evrose54

evrose54 (approved):
New updates to this branch (replaced ``str`` argument with ``string``) look good to me. Happy to have this merged in!

tests/unit_tests/commandline/test_log_setup.py
```diff
@@ -23,23 +22,53 @@ def generate_random_string(length):
...
+        string = string[:loc] + " " + string[loc + 1 :]
+        loc += random.randint(2, 8)
+    return string
+
+
 def generate_random_messages():
     """Generate a random amount of messages with random length."""
-    num_messages = random.randint(1, 50)
-    return [generate_random_string(random.randint(5, 110)) for _ in range(num_messages)]
+    num_messages = 20
```
jsolbrig, line 51: Thank you for making this a fixed number of tests. That will fix a lot of confusion.

Is there any reason that 20 wouldn't be a sufficient number of tests?

tests/unit_tests/commandline/test_log_setup.py
```diff
@@ -23,23 +22,53 @@ def generate_random_string(length):
...
     caplog.set_level(logging.INFO)
-    max_message_len = min(74, len(message))
-    assert max_message_len <= 80, "Max emphasis in '*' is longer than 80 chars."
     log_with_emphasis(LOG.info, message)
-    assert "*" * (max_message_len + 6) in caplog.text
-    for wmessage in wrap(message, width=74):
-        assert "** " + wmessage in caplog.text
-    assert "*" * (max_message_len + 6) in caplog.text
+    assert (  # top/bottom of box is formmated correctly
+        "*" * 9  # three for boarders, and min of 5 for string length
```
jsolbrig, line 67: borders, not boarders

jsolbrig, line 67: I appreciate the simplified test here. 

Feel free to explain why this isn't needed. Maybe it should test the length of all of the lines to ensure that they are equal unless there is no space in the line? Something like this?

```
stripped = [line.strip('**').strip() for line in lines]
stripped_lens = [len(line) for line in stripped if " " not in line]
assert len(set(stripped_lens)) == 1
```

tests/unit_tests/commandline/test_log_setup.py
```diff
@@ -23,23 +22,53 @@ def generate_random_string(length):
...
+    )
+    assert "** " in caplog.text
+    assert " **" in caplog.text
```
jsolbrig, line 70: Maybe use `caplog.text.startswith()` and `caplog.text.endswith()` here?

tests/unit_tests/commandline/test_log_setup.py
jsolbrig, line 1: It looks like this doesn't test long filenames. Should it or is that not necessary?

tests/unit_tests/commandline/test_log_setup.py
biosafetylvl5: Yeah, unfortunately that's not possible because the logging prefix is in the log text. :/ Theoretically, we could filter the log prefix, but that would need updating any time we change the log format

biosafetylvl5: this IS possible with caplog.messages - added 😄

tests/unit_tests/commandline/test_log_setup.py
biosafetylvl5: Added!

tests/unit_tests/commandline/test_log_setup.py
biosafetylvl5: Thanks 😭  Fixed.

tests/unit_tests/commandline/test_log_setup.py
jsolbrig: Okay, I think this sounds fine to me. If we start having random failures in this, though, we may want to specifically add a few of the edge cases. Things like specific special characters that we expect could cause problems.

I don't see any reason that we would actually expect to have issues with any edge cases, though.

tests/unit_tests/commandline/test_log_setup.py
jsolbrig: Great, thanks!

biosafetylvl5:
All relevant unit tests and all base tests passing - merging this now! 🎉

Issue #451, opened 2024-03-14
Provide a more informative error message when --compare_path is used without installing testing dependencies


evrose54:
# Requested Update

### Description
Stemming from [this comment](https://github.com/NRLMMD-GEOIPS/geoips/pull/450/files#r1525371612) from this [PR](https://github.com/NRLMMD-GEOIPS/geoips/pull/450), we've came to agreement that we need a better error message when ``--compare_path``, and any other GeoIPS testing functionality is used without the proper testing dependencies installed. 

### Background and Motivation
Currently, when using a fresh installation of geoips (new conda environment,`` pip install $GEOIPS_PACKAGES_DIR/geoips``), and then running ``geoips/tests/integration_tests/base_test.sh``, one of the tests will fail with the following error:

```
Traceback (most recent call last):
  File "/Users/erose/anaconda3/envs/geoips/bin/run_procflow", line 8, in <module>
    sys.exit(main())
  File "/Users/erose/geoips/geoips_packages/geoips/geoips/commandline/run_procflow.py", line 69, in main
    RETVAL = PROCFLOW(COMMAND_LINE_ARGS["filenames"], COMMAND_LINE_ARGS)
  File "/Users/erose/geoips/geoips_packages/geoips/geoips/plugins/modules/procflows/single_source.py", line 1923, in call
    retval += output_checker(
  File "/Users/erose/geoips/geoips_packages/geoips/geoips/plugins/modules/output_checkers/image.py", line 230, in call
    retval = plugin.compare_outputs(
  File "/Users/erose/geoips/geoips_packages/geoips/geoips/interfaces/module_based/output_checkers.py", line 875, in compare_outputs
    ) = self.perform_comparisons(
  File "/Users/erose/geoips/geoips_packages/geoips/geoips/interfaces/module_based/output_checkers.py", line 800, in perform_comparisons
    goodcomps, badcomps, compare_strings = test_product_func(
  File "/Users/erose/geoips/geoips_packages/geoips/geoips/interfaces/module_based/output_checkers.py", line 991, in test_products
    if self.module.outputs_match(
  File "/Users/erose/geoips/geoips_packages/geoips/geoips/plugins/modules/output_checkers/image.py", line 121, in outputs_match
    from pixelmatch.contrib.PIL import pixelmatch
ModuleNotFoundError: No module named 'pixelmatch'
```
[...]

PR #438, opened 2024-02-12
Clean up optional dependencies throughout GeoIPS Code

evrose54:
# Reviewer Checklist

* [x] Required ***existing tests*** pass (ie full_test.sh, others as appropriate)
* [x] NO REQUIRED ***unit tests*** (explain why not required)
* [x] NO REQUIRED ***integration tests*** (explain why not required)
* [x] Required ***documentation*** added for new/modified functionality
* [x] Required ***release notes*** added for new/modified functionality
* [x] Required ***updates to other repos*** complete
       - Data_fusion / geoips_clavrx need updates, see PR's shown below for those changes.

# Related Issues
fixes #338 
This data_fusion [PR](https://github.com/NRLMMD-GEOIPS/data_fusion/pull/30) fixes optional dependencies for that repo.
This geoips_clavrx [PR](https://github.com/NRLMMD-GEOIPS/geoips_clavrx/pull/34) fixes optional dependencies for that repo.

# Testing Instructions
You can run geoips/tests/integration_tests/full_test.sh for testing this PR. Tried it on my side and got a 0 on return. Tests pass.

# Summary
*From GEOIPS#338: 2023-07-19, Clean up optional dependencies*

Some GeoIPS readers include optional dependency statements that are required to read
certain file formats. While keeping these imports as optional is ok, we should clean up
the manner in which this is implemented. To do so, we've created a separate
``geoips.testing.context_manager.py`` script which can handle optional imports scattered
throughout the GeoIPS codebase. This is essentially replacing our old manner of optional
dependencies with a new method that keeps things clean.

    - modified: geoips/plugins/modules/procflows/config_based.py
    - modified: geoips/plugins/modules/readers/abi_netcdf.py
    - modified: geoips/plugins/modules/readers/ahi_hsd.py
    - modified: geoips/plugins/modules/readers/ewsg_netcdf.py
    - modified: geoips/plugins/modules/readers/modis_hdf4.py
    - modified: geoips/plugins/modules/readers/seviri_hrit.py
    - modified: geoips/plugins/modules/readers/utils/geostationary_geolocation.py
    - modified: geoips/plugins/modules/readers/viirs_netcdf.py
    - modified: geoips/utilts/memusg.py
    - added: geoips/utils/context_manager.py

PR #477, opened 2024-04-08
Provide better error message for --compare_path without testing dependencies

biosafetylvl5:
* [x] NO REQUIRED ***documentation*** (explain why not required)

# Related Issues
fixes NRLMMD-GEOIPS/geoips#451

# Testing Instructions
* Enter a `bash` environment in the Dockerfile, remove `pixelmatch`, run `tests/integration_tests/base_install.sh` and then `tests/integration_tests/base_test.sh`. `cat /output/logs/20240408/20240408.223821_geoips_base/test_all_geoips_base.log_amsr2_ocean.tc.windspeed.imagery_clean.sh.log` to see error message:

```bash
Processing complete! Checking outputs...


  File "/app/geoips_packages/geoips/geoips/plugins/modules/output_checkers/image.py", line 123, in outputs_match
    from pixelmatch.contrib.PIL import pixelmatch
ModuleNotFoundError: No module named 'pixelmatch'

The above exception was the direct cause of the following exception:

  File "/home/geoips_user/.local/bin/run_procflow", line 8, in <module>
    sys.exit(main())
  File "/app/geoips_packages/geoips/geoips/commandline/run_procflow.py", line 69, in main
    RETVAL = PROCFLOW(COMMAND_LINE_ARGS["filenames"], COMMAND_LINE_ARGS)
  File "/app/geoips_packages/geoips/geoips/plugins/modules/procflows/single_source.py", line 1917, in call
    retval += output_checker(
  File "/app/geoips_packages/geoips/geoips/plugins/modules/output_checkers/image.py", line 230, in call
    retval = plugin.compare_outputs(
  File "/app/geoips_packages/geoips/geoips/interfaces/module_based/output_checkers.py", line 849, in compare_outputs
    ) = self.perform_comparisons(
  File "/app/geoips_packages/geoips/geoips/interfaces/module_based/output_checkers.py", line 776, in perform_comparisons
    goodcomps, badcomps, compare_strings = test_product_func(
  File "/app/geoips_packages/geoips/geoips/interfaces/module_based/output_checkers.py", line 965, in test_products
    if self.module.outputs_match(
  File "/app/geoips_packages/geoips/geoips/plugins/modules/output_checkers/image.py", line 125, in outputs_match
    raise ImportError(
ImportError: Comparing images using GeoIPS requires a python library called pixelmatch (link: https://pypi.org/project/pixelmatch/).You can install it directly or install the test extra eg. pip install geoips[test]
```

# Summary
* Added a clear and informative error message when `--compare_path` is used without the pixelmatch dependency installed.
* Checked for other testing functionalities in `run_procflow` and added appropriate error messages for missing dependencies. See #451 for details.

# Output
* Command line output demonstrating the new error message when `--compare_path` is used without installing pixelmatch.
* No imagery output is required as the changes are related to error handling.


geoips/plugins/modules/output_checkers/image.py
```diff
@@ -118,7 +118,17 @@ def outputs_match(plugin, output_product, compare_product, threshold=0.05):
     )
     from PIL import Image
     import numpy as np
-    from pixelmatch.contrib.PIL import pixelmatch
+
+    try:
```
evrose54, line 122: We should wait on merging this PR until #438 is merged. Then, we should replace the optional import below with the new ``import_optional_dependencies`` functionality that has been created in #438.

jsolbrig: @biosafetylvl5 This comment is ready to be addressed.

biosafetylvl5: @evrose54 @jsolbrig The error message below has more details, and would be more useful to users. What benefit do we get by replacing with the `import_optional_dependencies` function? I'm so fine with doing so, but from what I see, it seems worse for the user. It also appears to fail to raise the error exception with traceback?

Let me know what I'm missing here 😵‍💫

jsolbrig: In this case I agree with you. Since you've implemented a better error message and this is a special case, I think we should keep this.

The history behind `import_optional_dependencies` is that there were dozens of these optional imports scattered all over the code, many with useless error messages. Maybe import_optional_dependencies wasn't the right solution, but it helped at least add something useful to those import errors to tell people what to do.

That said, the current error message still isn't very helpful and should be improved. We should think about how to improve the error message and how to get rid of as many optional imports as possible.

biosafetylvl5: Perhaps we can implement package specific messages, so it'd - for example - be applicable here, but still centralized. For now, I think this PR is gtg.

jsolbrig (approved):

mindyls (approved):
Looks good to me.

PR #494, opened 2024-04-15
37 geoips only install

mindyls:
Please see comments on "Files changed" tab.  A couple spots where if you only have geoips installed, things break on "sector_adjusters" (since no sector_adjuster plugins in geoips repo), or where there was a hard coded "geoips_clavrx" in the tests.

Note these updates were initially started when attempting to run "geoips test unit-test commandline" - we have since decided that unit test is intended to test the FULL geoips install, including all available plugin packages (since that is the only way to thoroughly test the CLI), but some of these changes I think are still applicable even if we are assuming the CLI unit test is exhaustive.

If this looks ok, we can just resolve the conversations here and merge directly to the 37-create-useful-geoips-CLI branch, before merging that one.

### DIFF
--- a/geoips/commandline/geoips_command.py
+++ b/geoips/commandline/geoips_command.py
@@ -294,9 +294,18 @@ class GeoipsExecutableCommand(GeoipsCommand):
               ("module_based", "yaml_based", "text_based")
         """
         if package_name == "all":
-            interface_registry = interface.plugin_registry.registered_plugins[
-                interface.interface_type
-            ][interface.name]
+            # If there are no plugins of current interface,
+            # just return None, do not fail catastrophically.
+            # This will fail on "sector_adjusters" interface
+            # during "geoips list plugins" if only geoips
+            # repo is installed (since there are no "sector_adjuster"
+            # plugins in the geoips repo)
+            if interface.name in interface.plugin_registry.registered_plugins[interface.interface_type]:
+                interface_registry = interface.plugin_registry.registered_plugins[
+                    interface.interface_type
+                ][interface.name]
+            else:
+                return None
         else:
             interface_registry = json.load(
                 open(resources.files(package_name) / "registered_plugins.json", "r")
--- a/tests/unit_tests/commandline/test_geoips_get_plugin.py
+++ b/tests/unit_tests/commandline/test_geoips_get_plugin.py
@@ -24,9 +24,16 @@ class TestGeoipsGetPlugin(BaseCliTest):
         if not hasattr(self, "_cmd_list"):
             self._cmd_list = []
             base_args = self._get_plugin_args
-            # validate all plugins from package geoips_clavrx
+            # validate all plugins from all packages
             for interface_name in interfaces.__all__:
                 interface = getattr(interfaces, interface_name)
+                # If there happen to be no plugins for a given
+                # interface, just skip, do not fail catastrophically.
+                # This allows defining interfaces in the geoips repo,
+                # even if there are no plugins of that interface
+                # directly in the main geoips repo.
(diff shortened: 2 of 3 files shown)


geoips/commandline/geoips_command.py
```diff
@@ -294,9 +294,18 @@ def _get_registry_by_interface_and_package(self, interface, package_name):
...
+            # This will fail on "sector_adjusters" interface
+            # during "geoips list plugins" if only geoips
+            # repo is installed (since there are no "sector_adjuster"
+            # plugins in the geoips repo)
+            if interface.name in interface.plugin_registry.registered_plugins[interface.interface_type]:
+                interface_registry = interface.plugin_registry.registered_plugins[
+                    interface.interface_type
+                ][interface.name]
+            else:
+                return None
```
mindyls, line 308: This caused "geoips list plugins" to fail - so probably the one we *do* want to fix.

tests/unit_tests/commandline/test_geoips_get_plugin.py
```diff
@@ -24,9 +24,16 @@ def all_possible_subcommand_combinations(self):
...
+            # validate all plugins from all packages
             for interface_name in interfaces.__all__:
                 interface = getattr(interfaces, interface_name)
+                # If there happen to be no plugins for a given
+                # interface, just skip, do not fail catastrophically.
+                # This allows defining interfaces in the geoips repo,
+                # even if there are no plugins of that interface
+                # directly in the main geoips repo.
+                if interface_name not in interface.plugin_registry.registered_plugins[interface.interface_type]:
+                    continue
```
mindyls, line 36: I don't think this will hurt anything - will just allow running on subsets of packages in the future (when we are not assuming geoips test unit-test commandline will run everything)

tests/unit_tests/commandline/test_geoips_validate.py
```diff
@@ -22,18 +22,19 @@ def all_possible_subcommand_combinations(self):
...
+                for plugin_type in ["modules", "yaml"]:
+                    if plugin_type == "modules":
+                        plugin_path_str = f"{pkg_path}/{plugin_type}/**/*.py"
+                    else:
+                        plugin_path_str = f"{pkg_path}/{plugin_type}/**/*.yaml"
+                    plugin_paths = sorted(glob(plugin_path_str, recursive=True))
+                    for plugin_path in plugin_paths:
+                        self._cmd_list.append(base_args + [plugin_path])
+                # Add argument list to retrieve help message
+                self._cmd_list.append(base_args + ["-h"])
```
mindyls, line 37: I am assuming this is what was intended for the "all_possible_subcommand_combinations - and not only running in the "geoips_clavrx" package?  Let me know if this was not the intent.

geoips/commandline/geoips_command.py
mindyls: Note these need black applied .... Sorry, can't push updates right now.

tests/unit_tests/commandline/test_geoips_validate.py
evrose54: I chose to run on geoips_clavrx because it is a fairly simple and small package. We should probably change the name from "all_possible_subcommand_combinations" to "selected_subcommand_combinations" or something along those lines. Some unit tests are stochastic in the command combinations they will test and we should make that variable reflect that ideology.

geoips/commandline/geoips_command.py
evrose54: It did? This does not fail on my end. What packages do you have installed?

evrose54:
I included these changes in #444, so we can probably just close this PR.

PR #497, opened 2024-04-16
fixed bugs introduced by finish log with emphasis pr

evrose54:
# Reviewer Checklist

* [x] Required ***updates to other repos*** complete (see NRLMMD-GEOIPS/recenter_tc#33)

# Related Issues
stems from bugs introduced by NRLMMD-GEOIPS/geoips#464
see NRLMMD-GEOIPS/recenter_tc#33

# Testing Instructions
Run ``pytest -v geoips/tests/unit_tests/commandline/test_log_setup.py`` and ``./geoips/tests/integration_tests/full_test.sh``

# Summary
*Finish log with emphasis PR (#464)* introduced a couple of bugs that weren't caught by
GeoIPS devs. This PR fixes those bugs and ensues that both the unit tests and accociated
scripts pass without failure. This fix changes a bug in test_log_setup.py where
``generate_random_messages(add_long_word=True)`` would return a ``map`` object rather
than an iterable list of strings. ``map`` objects have no length attribute and caused
failures within ``geoips/commandline/log_setup:log_with_emphasis``. We also added a new
check in ``log_with_emphasis`` that raises a ``TypeError`` if one or more of the items
in the unpacked list of ``*messages`` doesn't have a length attribute. Ie. None, map,
etc.

    modified: geoips/commandline/log_setup.py
    modified: tests/unit_tests/commandline/test_log_setup.py

### DIFF
--- a/geoips/commandline/log_setup.py
+++ b/geoips/commandline/log_setup.py
@@ -35,7 +35,7 @@ def log_with_emphasis(print_func, *messages):
     """
     wrapped_messages = []
     # Cast as a string just in case something other than a string exists in messages
-    messages = filter(lambda s: len(str(s)), messages)
+    messages = filter(lambda s: len(str(s)) > 0, messages)
     for message in messages:
         # wrap the message to a specified length
         wrapped_messages += wrap(message, width=74, break_long_words=False)
--- a/geoips/commandline/log_setup.py
+++ b/geoips/commandline/log_setup.py
@@ -34,8 +34,12 @@ def log_with_emphasis(print_func, *messages):
         The messages to be logged with emphasis
     """
     wrapped_messages = []
-    # Cast as a string just in case something other than a string exists in messages
-    messages = filter(lambda s: len(s) > 0 or len(str(s)) > 0, messages)
+    # Handle potential typeerror by casting as a string
+    try:
+        messages = filter(lambda s: len(s) > 0, messages)
+    except TypeError:
+        messages = filter(lambda s: len(str(s)) > 0, messages)
+
     for message in messages:
         # wrap the message to a specified length
         wrapped_messages += wrap(message, width=74, break_long_words=False)
--- a/geoips/commandline/log_setup.py
+++ b/geoips/commandline/log_setup.py
@@ -34,7 +34,12 @@ def log_with_emphasis(print_func, *messages):
         The messages to be logged with emphasis
     """
     wrapped_messages = []
-    messages = filter(lambda s: len(s) > 0, messages)
+    try:
+        messages = filter(lambda s: len(s) > 0, messages)
+    except TypeError:
+        raise TypeError(
+            f"1+ words sent to log_with_emphasis doesn't have a length attr. {messages}"
+        )
     for message in messages:
         # wrap the message to a specified length
         wrapped_messages += wrap(message, width=74, break_long_words=False)
--- a/geoips/commandline/log_setup.py
+++ b/geoips/commandline/log_setup.py
@@ -38,7 +38,7 @@ def log_with_emphasis(print_func, *messages):
         messages = filter(lambda s: len(s) > 0, messages)
     except TypeError:
         raise TypeError(
-            f"1+ words sent to log_with_emphasis doesn't have a length attr. {messages}"
+            f"1+ words sent to log_with_emphasis don't have a length attr. {messages}"
         )
     for message in messages:
         # wrap the message to a specified length
--- a/geoips/commandline/log_setup.py
+++ b/geoips/commandline/log_setup.py
@@ -34,11 +34,10 @@ def log_with_emphasis(print_func, *messages):
         The messages to be logged with emphasis
     """
     wrapped_messages = []
-    # Handle potential typeerror by casting as a string
-    try:
-        messages = filter(lambda s: len(s) > 0, messages)
-    except TypeError:
-        messages = filter(lambda s: len(str(s)) > 0, messages)
+    # Even though only strings should be passed here, we are going to allow any type of
+    # object to be passed for the time being. Best not to overthink the situation and
+    # just cast for the time being.
+    messages = filter(lambda s: len(str(s)) > 0, messages)
 
     for message in messages:
         # wrap the message to a specified length
--- a/geoips/commandline/log_setup.py
+++ b/geoips/commandline/log_setup.py
@@ -34,12 +34,8 @@ def log_with_emphasis(print_func, *messages):
         The messages to be logged with emphasis
     """
     wrapped_messages = []
-    try:
-        messages = filter(lambda s: len(s) > 0, messages)
-    except TypeError:
-        raise TypeError(
-            f"1+ words sent to log_with_emphasis don't have a length attr. {messages}"
-        )
+    # Cast as a string just in case something other than a string exists in messages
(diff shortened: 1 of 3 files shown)


tests/unit_tests/commandline/test_log_setup.py
```diff
@@ -151,6 +151,7 @@ def test_log_with_emphasis_long_word(message, caplog):
     log_with_emphasis(LOG.info, message)
     log_lines = caplog.messages[1 : len(caplog.text) - 1]
 
+    print(f"LOG LINES = {log_lines}")
```
biosafetylvl5: Are we wanting to leave this in?

tests/unit_tests/commandline/test_log_setup.py
```diff
@@ -84,9 +84,9 @@ def generate_random_messages(add_long_word=False):
     if add_long_word:
 
         def f(s):
-            insert_random_string_randomly(s, 88)  # insert string 88 chars long
+            return insert_random_string_randomly(s, 88)  # insert string 88 chars long
```
biosafetylvl5, line 87: Oh gosh! Can't believe this got through review. Thanks for fixing

geoips/commandline/log_setup.py
```diff
@@ -34,7 +34,12 @@ def log_with_emphasis(print_func, *messages):
         The messages to be logged with emphasis
     """
     wrapped_messages = []
-    messages = filter(lambda s: len(s) > 0, messages)
+    try:
+        messages = filter(lambda s: len(s) > 0, messages)
+    except TypeError:
+        raise TypeError(
+            f"1+ words sent to log_with_emphasis doesn't have a length attr. {messages}"
```
biosafetylvl5, line 41: Why add this?

biosafetylvl5 (commented):
Thanks @evrose54! I left some comments :)

tests/unit_tests/commandline/test_log_setup.py
evrose54: I removed that. Caught that right after I pushed

tests/unit_tests/commandline/test_log_setup.py
evrose54: No problem!

geoips/commandline/log_setup.py
evrose54: I was getting a TypeError in recenter_tc when I supplied an ``area_def`` in messages to ``log_with_emphasis``. All in all, this is very unlikely to occur, but will raise a more appropriate error if anything in ``*messages`` doesn't have a ``len`` attribute.

biosafetylvl5: Ah - is this when you pass something that isn't a string? Perhaps the filter should use 
```python
lambda s: len(s)>0 or len(str(s))>0
```
What do you think?

evrose54: Sure, that's probably a better solution. I'll make that update now.

evrose54: Updated!

biosafetylvl5 (approved):
Added an [or short circuit](https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not) for optimization since most messages will have a length property

geoips/commandline/log_setup.py
```diff
@@ -34,7 +34,8 @@ def log_with_emphasis(print_func, *messages):
         The messages to be logged with emphasis
     """
     wrapped_messages = []
-    messages = filter(lambda s: len(s) > 0, messages)
+    # Cast as a string just in case something other than a string exists in messages
+    messages = filter(lambda s: len(s) > 0 or len(str(s)) > 0, messages)
```
evrose54, line 38: I still think this will fail if any of the messages in ``*messages`` don't have a ``len`` attribute. For example, if a ``TypeError`` was raised when ``len(s)`` was called, then ``len(str(s))`` would never occur because the error was raised, correct?

biosafetylvl5: You're right. 😅 I think a try, except catch is probably the way then. I can add that

geoips/commandline/log_setup.py
```diff
@@ -34,7 +34,12 @@ def log_with_emphasis(print_func, *messages):
         The messages to be logged with emphasis
     """
     wrapped_messages = []
-    messages = filter(lambda s: len(s) > 0, messages)
+    # Handle potential typeerror by casting as a string
+    try:
+        messages = filter(lambda s: len(s) > 0, messages)
+    except TypeError:
+        messages = filter(lambda s: len(str(s)) > 0, messages)
+
```
jsolbrig, line 42: Is this correct? This doesn't appear to match the comment in the release notes. The release notes say:

> "We also added a new check in ``log_with_emphasis`` that raises a ``TypeError`` if one or more of the items
in the unpacked list of ``*messages`` doesn't have a length attribute."

But it looks like this masks these errors and allows them to pass. If this function were to receive `None`, it would enter the `except` and see a length of `4` because `len(str(None))` returns `4`. This would then add `None` to the set of messages.

jsolbrig: Also, is there a unit test that checks for appropriate behavior when one of the messages is `None`? What is the appropriate behavior? Should it ignore or raise TypeError?

jsolbrig (changes requested):
I think this needs to be looked at to ensure that cases like `None` are appropriately handled.

geoips/commandline/log_setup.py
evrose54: Sorry, @biosafetylvl5 took over part of this PR and I forgot to update the release notes to reflect the new changes made. To my knowledge, we chose to implement the code above as it is currently so we could handle data types that were not strings but could be printed in a string fashion (floats, ints, objects, etc.). However, I don't agree with None being a valid item that could be printed. Instead, should we do something along these lines?

```
try:
      messages = filter(lambda s: len(s) > 0, messages)
except:
      if None in messages:
          raise TypeError('some_error')
      messages = filter(lamda s: len(str(s)) > 0, messages)
```

jsolbrig: Do we have any situations where we already pass a non-string to this? If so, I'm fine with your solution. If not, I'd probably just limit it to strings.

Alternatively, we could just allow `None` to be printed too and let the developer who is calling the function sort that out.

I'll approve whichever solution you think is best here:
- Only allow strings
- Allow anything that isn't `None`
- Allow anything

evrose54: As you said in our meeting earlier, we shouldn't need to babysit developers using this functionality. I say we raise a TypeError if we encounter anything that is not the instance of a string. If you wan't ``log_with_emphasis`` to work, you *must* supply ``messages`` only as strings. So in this case, if you really wanted to print a specific object that is not a string, make sure to wrap it with ``str(object)`` before invoking ``log_with_emphasis``. How's that sound?

evrose54: After @jsolbrig and I spoke, we decided we are going to allow anything that is passed to ``log_with_emphasis``. We are now just going to preemptively cast every item in messages as a ``str`` object, so that we can print and check the ``len`` attribute of each object without failure. Developers should take the time to consider how they want ``non_string`` objects to be printed if they don't want a simple cast.

jsolbrig (approved):

Issue #632, opened 2024-06-05
Bug introduced in matplotlib versions ``>=3.9``


evrose54:
# Requested Update

### Description
In colormapper ``matplotlib_linear_norm``  we use a function call ``matplotlib.cm.get_cmap`` which no longer works. If you look at matplotlib's documentation for that function, it still exists, which is quite weird. Luckily this function exists in two locations so we can just refactor to the other use case, which is ``maplotlib.pyplot.get_cmap``. 

This [issue](https://github.com/matplotlib/matplotlib/issues/28349) posted by @jsolbrig explains it in more detail.

### Alternative Solutions
We can either:
* Downgrade to matplotlib versions ``<3.9``
* Change ``matplotlib.cm.get_cmap`` to ``matplotlib.pyplot.get_cmap``

### Code to demonstrate issue
``geoips.plugins.modules.colormappers.matplotlib_linear_norm:call``

### Checklist for Completion
* [ ] Replace all instances ``matplotlib.cm.get_cmap`` with ``matplotlib.pyplot.get_cmap``

PR #633, opened 2024-06-05
fixed bug introduced by matplotlib versions >=3.9

evrose54:
# Reviewer Checklist

* [x] NO REQUIRED ***documentation*** (explain why not required)
* [x] Required ***release notes*** added for new/modified functionalitye

# Related Issues
fixes NRLMMD-GEOIPS/geoips#632

# Testing Instructions
I only tested on ``./base_test.sh`` and will run ``./full_test.sh`` once I can find the time.

# Summary
Prior to matplotlib versions ``>=3.9`` we were able to use ``matplotlib.cm.get_cmap``
without any problem. After 3.9 was introduced, this failed because
``cm had no attribute called 'get_cmap'``. To fix this, we've migrated such calls from
``cm.get_cmap`` to ``pyplot.get_cmap``, as that function still works for pyplot. It's
weird that the same functionality was located in two different places, but at least it
makes for an easy fix.

    modified: geoips/image_utils/colormap_utils.py
    modified: geoips/plugins/modules/colormappers/matplotlib_linear_norm.py

### DIFF
--- a/docs/source/releases/v1_13_0a0.rst
+++ b/docs/source/releases/v1_13_0a0.rst
@@ -25,6 +25,7 @@ Version 1.13.0a0 (2024-04-24)
 * Bug fixes
 
   * Fix color table application in geotiff_standard output formatter
+  * Fix bug introduced by matplotlib versions ``>=3.9``
 
 Enhancements
 ============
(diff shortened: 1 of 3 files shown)


mindyls (approved):
Great, thanks for fixing this!!

jsolbrig (approved):

PR #635, opened 2024-06-05
634 incorrect capitalization on this error should be clierror

biosafetylvl5:
No change to functionality, just changing `Cli` to `CLI` in class names as per PEP8.


evrose54 (changes requested):
Looks good to me, however, if we're going to enforce this change would you mind going through the codebase and changing class names where this case occurs? I know for a fact that most (if not all) of the CLI Classes have this problem and there may be some other instances of this class naming issue throughout the code base. 

Might as well be consistent! Thanks for doing this @biosafetylvl5.

biosafetylvl5:
Closing; will not capitalize acronyms in class names and will instead use camel case

Issue #576, opened 2024-05-15
CLI: Reduce number of command layers where possible and appropriate


jsolbrig:
# Blocking issues
This is likely blocked by #573 or vice versa.

# Requested Update

### Description
There are several locations where the CLI is overly verbose because an extra command level is required. Here is a **non-exhaustive** list of examples:
- `geoips list plugins <interface_name>` -> `geoips list <interface_name>`
- `geoips get plugin <interface_name> <plugin_name>` -> `geoips get <interface_name> <plugin_name>`

We should find all instances where we can remove an intermediate command and still have the CLI make sense grammatically. `geoips list algorithms` makes just as much sense as `geoips list plugins algorithms`. In some cases, this will require adding an additional command for listing everything. For example, if we remove `geoips list plugins` in favor of `geoips list <interface_name>` we might want to add `geoips list all-plugins` or something similar.

Note that the examples given here all relate to `<interface_name>`. There may be other examples of subcommands that can be moved upwards.

### Background and Motivation
This will reduce verbosity and makes the CLI feel more natural. It is a bit more like Kubernetes which seems to be a well thought-out CLI.

### Checklist for Completion
* [ ] Find and list in this issue all locations where we can and should remove verbosity.
* [ ] Remove the appropriate command level and replace with the lower-level commands.
* [ ] Determine where "all" commands should be added and add them where appropriate (e.g. `geoips list all-plugins`).
* [ ] Ensure that unit tests continue to pass.
* [ ] Update documentation as appropriate, auto-generating where possible.

Issue #573, opened 2024-05-15
Use class factories to generate subcommands for commands like `geoips get family` and `geoips get <interface_name>`


jsolbrig:
# Blocking issues
This is likely blocked by #576 or vice versa.

# Requested Update

### Description
Use class factories to generate new subcommands where appropriate within the CLI. There are several locations where this could be done. Some examples include:

- `geoips list families <interface_name>`
- `geoips get family <interface_name> <family_name>`
- `geoips list plugins <interface_name>`
- `geoips get plugin <interface_name> <plugin_name>`

Where the interface_name would become a new set of dynamically created subcommands.

### Background and Motivation
Currently there are several locations where we call `add_arguments` with `choices` to handle the various interfaces. Commands like `geoips list interface <interface_name>` set `add_argument(choices=<list_of_interface_names>)`. Making these into subcommands rather than an argument with choices will make the CLI function more cleanly and consistently and will help with exposing information about the available commands from the CLI. It will also simplify usage messages.

### Alternative Solutions
Wait until we have made a decision about whether to change to an inheritance-based model for these classes. I'd prefer to do it now, though.
  
### Checklist for Completion
* [ ] Develop a class factory that can generically be used for adding subcommands to any parent command. At minimum, it should accept the parent command instance, a list of the subcommand names, and a function that will act as the `add_arguments` function on the new classes. Please implement any other required arguments to the class factory (e.g. other properties or functions that should be added via the factory.
* [ ] Find and list all locations where a class factory would be appropriate for generating subcommands.
* [ ] Run the unit tests to ensure that everything still works as before. This shouldn't really change behavior except in the help and usage messages.

PR #642, opened 2024-06-11
CLI class factory

evrose54:
# Reviewer Checklist

* [ ] Required ***unit tests*** modified for new/modified functionality

# Related Issues
partially addresses  NRLMMD-GEOIPS/geoips#573
partially addresses  NRLMMD-GEOIPS/geoips#576

# Testing Instructions
Run pytest -v ./tests/unit_tests/commandline/

# Summary
While this PR doesn't fix all of the problems addressed in the issues above, it is our
initial attempt at creating a class factory for certain CLI commands to reduce the
verbage needed to execute those commands. Spefically in this branch, we've addressed the
command class ``GeoipsListSingleInterface``. We now use the aforementioned class as a
base class to build ``GeoipsListSingleInterface<interface_name>`` classes at runtime.

By doing so, we can now run ``geoips list <interface_name>`` rather than
``geoips list interface <interface_name>``. This improves the readability of this
command and makes it much easier to type via the CLI. We expect in future PRs to address
similar commands, such as ``geoips get family <family_name>`` and
``geoips get plugin <interface_name> <plugin_name>`` for the reasons mentioned
previously.

    modified: docs/source/userguide/command_line.rst
    modified: geoips/commandline/ancillary_info/cmd_instructions.yaml
    modified: geoips/commandline/geoips_list.py
    modified: tests/unit_tests/commandline/cli_top_level_tester.py
    modified: tests/unit_tests/commandline/test_geoips_list_interface.py


tests/unit_tests/commandline/cli_top_level_tester.py
```diff
@@ -23,7 +24,7 @@ class BaseCliTest(abc.ABC):
     _get_interface_args = ["geoips", "get", "interface"]
```
jsolbrig, line 24: I think I'd like to move these onto each class. These are confusing where they are, though I do see how they simplify things a bit.

jsolbrig (dismissed):
Once the comment is addressed, this is good to go IMO.

tests/unit_tests/commandline/cli_top_level_tester.py
evrose54: I moved these on each class in #644

docs/source/releases/v1_13_0a0.rst
```diff
@@ -336,6 +337,34 @@ implement those changes.
...
+*From GEOIPS#576: 2024-05-11, CLI: Reduce number of command layers where possible and appropriate*
+*From GEOIPS#573: 2024-05-11, Use class factories to generate subcommands for commands like ``geoips get family`` and ``geoips get interface``
+
+While this PR doesn't fix all of the problems addressed in the issues above, it is our
+initial attempt at creating a class factory for certain CLI commands to reduce the
+verbage needed to execute those commands. Spefically in this branch, we've addressed the
+command class ``GeoipsListSingleInterface``. We now use the aforementioned class as a
+base class to build ``GeoipsListSingleInterface<interface_name>`` classes at runtime.
+
+By doing so, we can now run ``geoips list <interface_name>`` rather than
```
mindyls, line 352: Will this always be unique ? Ie, there will not be similarly named interfaces, plugins,  families that would all come after "geoips list" ? Or would it always be "geoips list <interface name> ..."?

geoips/commandline/geoips_command.py
```diff
@@ -38,11 +37,12 @@ def __init__(self):
         get_plugin_packages() and get_plugin_package_paths() functions.
         """
         self.entrypoints = [
-            ep.value for ep in sorted(get_entry_point_group("geoips.plugin_packages"))
+            ep.value
+            for ep in sorted(metadata.entry_points(group="geoips.plugin_packages"))
```
mindyls, line 41: Is this in another PR?

mindyls (approved):

geoips/commandline/geoips_command.py
evrose54: Yes, this is in v1.13 release PR however I needed this functionality in this branch otherwise the CLI would not work. This change will be removed once v1.13 is released. If this is merged first, it should not cause conflicts.

docs/source/releases/v1_13_0a0.rst
evrose54: I think this will always be unique. I don't think we'll ever support two interfaces that are named the same thing and retrieving plugins is handled at an interface by interface basis, so we only require that plugin names be unique within a given interface. I think that answers your question, but I'm not sure.

evrose54:
We should ignore ``Old docs have not been changed`` CI check as we need this documentation and will likely move all CLI documentation over to new docs in a single PR.

geoips/commandline/geoips_command.py
```diff
@@ -532,3 +532,75 @@ def add_arguments(self):
     def __call__(self, args):
         """Exectutable function for GeoipsListCommon Class. Not implemented."""
         pass
+
+
+class CommandClassFactory:
```
jsolbrig, line 537: I don't think we should change this now because there are dependent PRs, but I think this should just be a function rather than a class. It is used in a context where it is just used once, then thrown away, so the statefulness doesn't gain us anything and makes it a little unintuitive.

I don't think the change will be difficult. It would just be a matter of removing `self` and returning `classes` rather than accessing them via an attribute.

geoips/commandline/geoips_list.py
```diff
@@ -667,8 +665,11 @@ class GeoipsList(GeoipsCommand):
     """Top-Level List Command for listing off GeoIPS Artifacts."""
 
     name = "list"
-    command_classes = [
+    generated_classes = CommandClassFactory(
         GeoipsListSingleInterface,
+        interfaces.__all__,
+    ).classes
```
jsolbrig, line 671: If we make my other suggested update in a future issue, we would also need to change this.

jsolbrig (approved):

geoips/commandline/geoips_command.py
jsolbrig: See #708

geoips/commandline/geoips_list.py
jsolbrig: See #708

jsolbrig (approved):
This looks great! I mentioned one change in my comments but that change can be made after we get the remaining CLI PRs merged. See #708.

geoips/commandline/geoips_command.py
evrose54: Thanks for the high praise! Thinking about this after writing it; I agree that it should just be a function rather than a class. I'll make that update at a later point. Thanks for reviewing and creating that issue. Resolving for now.

geoips/commandline/geoips_list.py
evrose54: Sounds good! I'll make sure to implement what you recommended after this PR and its dependents get merged in. Resolving for now.

docs/source/releases/latest/575-cli-class-factory.yaml
```diff
@@ -0,0 +1,27 @@
...
+    verbage needed to execute those commands. Spefically in this branch, we've addressed the
+    command class ``GeoipsListSingleInterface``. We now use the aforementioned class as a
+    base class to build ``GeoipsListSingleInterface<interface_name>`` classes at runtime.
+
+    By doing so, we can now run ``geoips list <interface_name>`` rather than
+    ``geoips list interface <interface_name>``. This improves the readability of this
+    command and makes it much easier to type via the CLI. We expect in future PRs to address
+    similar commands, such as ``geoips get family <family_name>`` and
+    ``geoips get plugin <interface_name> <plugin_name>`` for the reasons mentioned
+    previously.
```
mindyls, line 17: @evrose54 @jsolbrig Wait, what happens if you have a plugin named the same thing in 2 different interfaces?  That is not out of the question?   How do you disambiguate?

evrose54: That's fine! In pr #644, we implement class factories for the ``geoips get command``. For example, if we had a plugin named ``Cloud-Height`` in two different interfaces, this is how it would work:
```bash
geoips get algorithm Cloud-Height
geoips get reader Cloud-Height
geoips get product-default Cloud-Height
...
```
You're forced to provide the interface name. This gives us enough information to find the correct plugin.

jsolbrig: It would look like this:
```
geoips list algorithms infrared
geoips list colormaps infrared
```

evrose54: I will say that ``list`` commands don't list specifically named plugins. If we were to do ``geoips list algorithms``, the only other arguments you could add are ``-p <package_name>``, ``--columns``, or ``--long``.

mindyls: mindy
  1:16 PM
I am playing with it a bit. I'm not sure how I feel with kind of lumping the "geoips list <interface_name>" and "geoips list <explicit_sub_cmd>" together under the same "list" command.  But I am leaning towards just merging it, and letting more people play with it, and getting feedback ?




NEW

[1:17](https://geoips.slack.com/archives/C0630KQ0KF0/p1721938643004979)
Because, at the same time, "geoips list interface algorithms" is also a "different" way of calling geoips list.  So either way is lumping 2 different types of calls to "geoips list" together, and I suppose fewer characters to type is as good of a reason to pick one over the other as any

mindyls (approved):
As long as @biosafetylvl5 is ok with the old docs being updated, I'm good to merge. (ie, this change is not going to get lost in the transition to new docs)

Issue #637, opened 2024-06-06
Using ``parser.parse_known_args`` in the CLI to apply scope to arguments


evrose54:
# Requested Update

### Description
There are a couple of instances in the GeoIPS CLI where we'd like to apply some scope to arguments available to a certain command but is currently not implemented, or if it is, it's done in a hacky way. For example, we'd like to share arguments amongst all commands for ``--log-level``, and possibly some other arguments that would apply to any command. There are also instances where we'd like to share common arguments amongst a group of commands. Currently, we share arguments amongst ``list`` commands via a ``GeoipsListCommon`` class. We add arguments to this class, and when initializing all list commands such as ``GeoipsListPackages``, ``GeoipsListPlugins``, etc. we add ``GeoipsListCommon`` as it's parent so they all inherit those shared arguments.

This likely could be made simpler by making use of ``argparse.ArgumentParser.parse_known_args`` command. This will allow parsing arguments that are known at the time of execution rather than at the end of initialization. By using this in a clever way, we can add global arguments to all commands and shared arguments by the group of commands. This way we can more easily control the scope of a command argument.

### Background and Motivation
This issue stems from these two comments ([1](https://github.com/NRLMMD-GEOIPS/geoips/pull/465/files/eb8b2b9e8bedab5ef5ed33105549d2ca8d946c33#r1619204653) : [2](https://github.com/NRLMMD-GEOIPS/geoips/pull/465/files/c41a37f9d01b59294c3d04917e9b80549364bca6#r1619182932)) on PR #465.

### Code to demonstrate issue
*  ``geoips.commandline.geoips_command:GeoipsCommand:__init__``
* ``geoips.commandline.geoips_command:GeoipsListCommon``

### Checklist for Completion
* [ ] Use ``argparse.ArgumentParser.parse_known_args`` to:
  * [ ] Add global arguments to all commands
  * [ ] Add command specific shared arguments
* [ ] Remove ``GeoipsListCommon`` class and the functionality that makes use of it

PR #659, opened 2024-07-12
Apply Scope to CLI Arguments

evrose54:
# Reviewer Checklist

# Related Issues
fixes NRLMMD-GEOIPS/geoips#637

# Testing Instructions
Run ``pytest -v $GEOIPS_PACKAGES_DIR/geoips/tests/unit_tests/commandline/`` and get no failed tests.

# Summary
While this PR ended up not making use of ``parse_known_args`` (It was buggy and
resulted in overly complex conditionals), we did end up finding a way to apply scope
(ie. share arguments from parents to children) to CLI commands to reduce repeated
portions of the code. This is also nice because we only have to make code changes to one
place if we want to alter arguments that are shared by various commands.

To do this, we created a ``ParentParsers`` Object in
``geoips.commandline.geoips_command`` which contains argument parsers that add arguments
which will be shared by some, if not all of the child command classes. For example, the
``geoips_parser`` attribute of ``ParentParsers`` will be shared to all child commands.
This is because every command is a child of ``geoips`` (the name of ``GeoipsCLI`` class).
A similar ideology applies to ``list_parser``, except that only children of ``list``,
ie. ``GeoipsListPackages``, ``GeoipsListPlugins``, etc. will get the arguments created
by ``list_parser``. This allows for us to share universal arguments such as log level
and command specific argumetns such as ``package_name``.

    modified: geoips/commandline/commandline_interface.py
    modified: tests/unit_tests/commandline/cli_top_level_tester.py


geoips/commandline/commandline_interface.py
```diff
@@ -76,6 +76,13 @@ def __init__(self, instructions_dir=None, legacy=False):
     def execute_command(self):
         """Execute the given command."""
         self.GEOIPS_ARGS = self.parser.parse_args()
+        # NOTE: We should discuss how we want to share the selected LOG to child classes
+        # They can all use the functionality below, however that would be redundant and
+        # there is likely a better way to fix this. Unfortunately 'parse_known_args'
+        # didn't work and this is our best solution for the time being.
+
+        # self.LOG = getattr(LOG, self.GEOIPS_ARGS.log_level)
+        # self.LOG("LOG LEVEL = {self.GEOIPS_ARGS.log_level}")
```
jsolbrig, line 85: Is there an issue for this? If not, we should create it.

geoips/commandline/geoips_command.py
```diff
@@ -52,6 +52,58 @@ def __init__(self):
...
+        "--columns",
+        "-c",
+        type=str,
+        nargs="+",
+        default=None,
+        help="""Specific Headers of Data you'd like to see listed.
+                For more in formation on headers available, run
+                'geoips list <cmd> <positional_args> --columns help'.
+                """,
+    )
```
jsolbrig, line 104: I haven't really seen someone use a class like this before. It's interesting! I think this is good for now, but I think we should consider two things:

1. Is there a better container for this than a class?
2. Is there a way to generalize this? Currently this will work only for the top-level commands and it would be difficult to apply propagating arguments to lower-level commands.

I'm running out of time today, but we should write an issue for this.

jsolbrig (approved):

geoips/commandline/geoips_command.py
```diff
@@ -52,6 +52,58 @@ def __init__(self):
...
+
+    geoips_parser = argparse.ArgumentParser(add_help=False)
+    geoips_parser.add_argument(
+        "--log_level",
+        "-log",
+        type=str,
+        default="interactive",
+        choices=["debug", "error", "info", "interactive", "warning"],
+        help="The logging level to use for output via the CLI.",
+    )
```
jsolbrig, line 75: I don't see `--log-level` in the output from `geoips -h`.

```
> geoips -h
usage: geoips [-h] {config,get,list,run,test,validate} ...

positional arguments:
  {config,get,list,run,test,validate}
                        geoips instructions.
    config              Various configuration-based commands for setting up your geoips environment. Currently supports `geoips config install <test_dataset_name>` command.
    get                 Retrieve an appropriate GeoIPS artifact. Currently supported `get` calls are: ["family", "interface", "package", "plugin"]
    list                List off available GeoIPS artifacts provided further commands. Current artifacts available for listing are: ["interface", "interfaces", "packages", "plugins", "scripts", "test-datasets", "unit-tests"].
    run                 Run a specified geoips process workflow (procflow) plugin. See 'geoips run single_source' or 'geoips run data_fusion' for more information of what arguments to providel
    test                Run tests implemented in GeoIPS or a separate GeoIPS Package. This could be running a set of linters on a specific package, running a certain integration or normal test script, or running unit tests. Current options are ["linting", "script", "unit-test"].
    validate            Validate a GeoIPS Plugin found at <file_path>. While this is done under the hood via GeoIPS, this is an easy way to test whether or not the plugin you are developing is valid.

options:
  -h, --help            show this help message and exit
```

I do see it in the output from `geoips list -h`, though.

jsolbrig (changes requested):
Create issues and address problem with `geoips --log-level`.

docs/source/releases/v1_14_0a0/0-pre-existing-conditions.yaml
jsolbrig, line 1: Need to update to `docs/source/releases/latest/`

geoips/commandline/geoips_command.py
evrose54: Addressing this now. Easy fix.

docs/source/releases/v1_14_0a0/0-pre-existing-conditions.yaml
evrose54: Fixed!

geoips/commandline/commandline_interface.py
evrose54: I'll create that issue.

evrose54: https://github.com/NRLMMD-GEOIPS/geoips/issues/714

geoips/commandline/geoips_command.py
evrose54: See #715.

mindyls (approved):
Agree with Jeremy's comments

jsolbrig (approved):

PR #688, opened 2024-07-18
Update to use latest instead of versions

biosafetylvl5:
Moved the yaml release files from `docs/source/release/v(version number)/*` to `docs/source/release/latest/*`, which now builds to `latest.rst`.

Added `latest.rst` because brassy does not automatically build .rst file at the moment. Ideally, brassy would create `latest.rst`, which could be pulled down and built into the docs locally. Without latest.rst the docs will not build. So adding a blank file serves as a placeholder until the CI automatically builds _and_ commits a latest.rst file. 

Added `latest` to `/docs/source/release/index.rst` so docs build.

The release note not edited check SHOULD NOT PASS, because.... it was edited 😄 

## Relevant XKCD:

CI be like:
image

### DIFF
--- a/docs/source/releases/latest/687-change-release-note-yaml-folder-from-version-to-latest.yaml
+++ b/docs/source/releases/latest/687-change-release-note-yaml-folder-from-version-to-latest.yaml
@@ -1,5 +1,5 @@
 continuous_integration:
-  description: |
+- description: |
     Moved the yaml release files from ``docs/source/release/v(version number)/*`` to ``docs/source/release/latest/*``, which now builds to ``latest.rst``.
     Added ``latest.rst`` because brassy does not automatically build .rst files at the moment. 
     Ideally, brassy would create ``latest.rst``, which could be pulled down and built into the docs locally.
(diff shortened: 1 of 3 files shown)


jsolbrig (approved):

docs/source/releases/latest/687-change-release-note-yaml-folder-from-version-to-latest.yaml
```diff
@@ -0,0 +1,15 @@
...
+    deleted:
+    - ''
+    modified:
+    - ''
+    moved:
+    - ''
+  related-issue:
+    number: 0
+    repo_url: ''
+  title: 'test'
```
mindyls, line 15: Does this need to actually be filled out?

docs/source/releases/latest.rst
mindyls, line 1: So we are starting with an empty version of latest.rst so that the doc build will pass?  Then this will just automatically be updated in the background with every PR, but it won't cause conflicts because it will be automatically updated AFTER finalizing and testing the PR?

mindyls: Yes, you explained this in your PR contents...

mindyls (dismissed):
Once the release notes are filled out, good to go.

docs/source/releases/latest/687-change-release-note-yaml-folder-from-version-to-latest.yaml
biosafetylvl5: Yes! Filled out now

mindyls (approved):
Thanks!

jsolbrig (approved):

PR #693, opened 2024-07-22
WORKSHOP BUG FIX pip install package from build_docs.sh if it is not already installed

mindyls:
This will ensure the plugin packages pass CI, since currently the build docs action does NOT pip install the package before calling build_docs.sh.

In the future we may not want to  support this, but just exit 1 if the package is not installed.  But for now, check if the package is installed, and pip install -e it if it is not installed already.

### DIFF
--- a/docs/build_docs.sh
+++ b/docs/build_docs.sh
@@ -69,6 +69,13 @@ if [[ ! -d "$docbasepath/$pkgname" ]]; then
     exit 1
 fi
 echo "package path=$docbasepath/$pkgname"
+retval=`python -c "import $pkgname"`
+if [[ "$retval" != "0" ]]; then
+    echo "***************************************************************************"
+    echo "ERROR: Package $pkgname is not installed"
+    echo "For now, pip installing $docbasepath, in future will fail"
+    pip install -e $docbasepath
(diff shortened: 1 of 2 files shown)


docs/build_docs.sh
```diff
@@ -69,6 +69,18 @@ if [[ ! -d "$docbasepath/$pkgname" ]]; then
...
+    echo "***************************************************************************"
+    echo "ERROR: Package $pkgname is not installed"
+    echo "For now, pip installing $docbasepath, in future will fail"
+    pip install -e $docbasepath
+    echo "***************************************************************************"
+else
+    echo "***************************************************************************"
+    echo "Package $pkgname is already installed!"
+    echo "***************************************************************************"
+fi
```
mindyls, line 83: Do not merge until pointing to dev.

christopher-selman (approved):

evrose54 (approved):

christopher-selman (approved):

jsolbrig (approved):
I'm approving this for now. It works. I also created #696 to move the installation from `build_docs.sh` to the CI but will deal with that later.

PR #697, opened 2024-07-22
WORKSHOP BUG FIX: Ensure return value correctly captured from import package within build_docs.sh

mindyls:
Required to fix GEOIPS/geoips#666

Was reinstalling every single time... Because I did not actually capture the return value...


jsolbrig (approved):
Hah, yes, this looks better!

PR #703, opened 2024-07-23
Improve error message for bad python version check_system_requirement…

biosafetylvl5:
When installing GeoIPS on a machine with only python3 (and not python) we get `WARNING: 'python --version' failed, please install python >= 3.9 before proceeding` because the code looks for `python` explicitly. Some Linux distributions decided during the transition from Python 2 to Python 3 that python should always refer to Python 2, and the command to run Python 3 would be python3 with a 3 at the end. Debian and Ubuntu did this. This change adds a message to users informing them how they can change their system so calls to `python` are re-routed to their local `python3`.

Relevant XKCD:


mindyls (approved):
Looks good to me.

jsolbrig (approved):

PR #730, opened 2024-08-06
Added generic_satpy reader plugin. 

### LINKED ISSUE #729: Add Generic Satpy Reader Plugin
# Requested Update

### Description
*  Add a GeoIPS wrapper for Satpy's reader functionality.

### Background and Motivation
*  Satpy provides a wide array of readers that GeoIPS could make use of.
* Pre-existing algorithms that rely on Satpy may be more easily ported to GeoIPS if a generic Satpy wrapper were available.

### Checklist for Completion
* [x] Write basic wrapper plugin
* [ ] Review code to ensure maximum compatibility with Satpy's readers 
* [ ] Determine testing strategy to ensure plugin continues to work with future versions of Satpy

A basic wrapper plugin is provided.  The plugin exposes all of the arguments required by a Scene constructor and the resulting Scene's load method to the GeoIPS command line interface via the reader_kwargs argument.  The plugin converts the channels loaded by the Scene object to a Dictionary of Datasets as required by the GeoIPS reader interface.  A dictionary passed via the reader_kwargs is used to group channels by resolution into the output Dictionary. The code assumes that a Scene's start_time and end_time attributes are available for every dataset in order to provide this information in the output Dictionary. 

The code has been used almost exclusively with a small amount of GOES16/18 data and Himawari data.  The code should be more systematically tested and evaluated.  Regression tests for this plugin may be particularly challenging to write and maintain since it will need as many tests as there are Satpy readers.  It may make sense to select a subset of readers to support and list the others as "use at your own risk."

robertdemariacira:
# Reviewer Checklist

**Remove lines that are not applicable, explain if you select "NO REQUIRED"**

* [x] NO REQUIRED ***updates to other repos*** (This adds a reader plugin to the GeoIPS repo and should not need to update any other repo.)

# Related Issues
fixes NRLMMD-GEOIPS/geoips#729

# Testing Instructions

At the moment, only cursory manual tests have been performed and no automated tests exist.  I believe automated tests should be added. However, given that this wrapper potentially adds 100+ new readers to GeoIPS, a strategy should be found to effectively test this functionality or prioritize test reading some data over others.

# Summary

Version X.Y.Z (YYYY-MM-DD)
**************************

* Added generic Satpy reader.

Major New Functionality
=======================

-------------------------------------------------

*From NRLMMD-GEOIPS/geoips#729: 2024-08-05*

* Adds generic Satpy reader plugin named generic_satpy
* Exposes arguments to Satpy Scene constructor and Scene load method to GeoIPS CLI via `reader_kwargs`
* Uses `channel_groups` argument to map loaded Scene to output Dictionary of Datasets

::

    added: geoips/plugins/modules/readers/generic_satpy.py

# Output


geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,134 @@
...
+    if load_kwargs is None:
+        load_kwargs = {}
+
+    scene.load(
+        all_channels,
+        calibration=calibration,
+        resolution=resolution,
+        polarization=polarization,
+        level=level,
+        modifiers=modifiers,
```
djhoese, line 96: It's possible that these 5 kwargs can be removed. They are used as filters to limit what channels are actually loaded. Or rather, what version of a channel is loaded. For example, if you were loading MODIS L1B data but you wanted the 1km version of all channels even though the 500m of band 1 and 2 was available, you would provide "resolution=[500]" (I think) along with channels being ["1", "2"]. I'm not sure GeoIPS would be able to usefully use this with the dataset naming/grouping later on in this function.

Similarly (alternatively?), it may make sense to first get a list of all DataIDs that the Scene could produce using `scn.available_data_ids()` and then removing any where `data_id["name"]` does not exist in `all_channels`. This would make the use of these 5 filter parameters more useful as this implementation I just described would be loading (potentially) multiple resolutions or calibrations of the same channel. But that would also be needed if someone wanted ABI "C14Rad" versus "C14BT".

geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,134 @@
...
+        generate=generate,
+        unload=unload,
```
djhoese, line 98: I don't think `unload` is useful to GeoIPS. Setting it to `False` would tell satpy's Scene to not clear out loaded datasets that aren't needed anymore now that a composite was created with them. I can't see a reason for GeoIPS to use this beyond debugging in which case you'd probably be modifying this call anyway.

geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,134 @@
...
+        load_kwargs = {}
+
+    scene.load(
+        all_channels,
+        calibration=calibration,
+        resolution=resolution,
+        polarization=polarization,
+        level=level,
+        modifiers=modifiers,
+        generate=generate,
```
djhoese, line 97: The `generate` kwarg controls whether or not composites are generated or rather *attempted* to be generated. So if `true_color` from Satpy can be made without any interpolation/resampling then we'd want `generate` to be `True` (the default). Unless this reader adds a `Scene.resample` call to do the interpolation on the satpy-side then `generate` should probably always be `True`.

This does bring up the bigger question though: should this reader allow users to load Satpy composites? In the cases where interpolation isn't needed then it would work fine, but in the cases where it is needed then you need a `Scene.resample` call. Otherwise, GeoIPS would need some major hand-holding added to produce all the dependencies for a Satpy composite (`generate=False`) and then recreate the `Scene` in a composite/product/algorithm step and then call `Scene.generate_possible_composites(True)` I think.

geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,134 @@
...
+    output_datasets = {}
+    for group_name, channel_names in channel_groups.items():
+        lons, lats = (None, None)
+        is_first_channel = True
+        data_dict = {}
+        for channel_name in channel_names:
+            channel = scene[channel_name]
+            if is_first_channel:
+                is_first_channel = False
+                lons_np, lats_np = channel.attrs["area"].get_lonlats()
```
djhoese, line 111: `get_lonlats()` is pretty inconsistent in what it returns. For `SwathDefinition`s like those created for VIIRS and other LEO/polar instruments, Satpy will probably be providing the swath definition an xarray DataArray but `get_lonlats()` will give you the dask array underneath. For `AreaDefinition`s it would be numpy arrays by default I think. You could pass `chunks=...` to this method and then you'd always get dask arrays and then could compute them together with `lons_np, lats_np = dask.array.compute(lons_dask, lats_dask)` or not compute them if you want GeoIPS to deal with the dask arrays later on.

geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,134 @@
...
+                lons = xr.DataArray(lons_np, dims={"y": channel.y, "x": channel.x})
+                lats = xr.DataArray(lats_np, dims={"y": channel.y, "x": channel.x})
```
djhoese, line 113: I would avoid the `.<attr>` access of coordinates in `DataArray`s and instead be explicit with `channel.coords["x"]` and `channel.coords["y"]`.

geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,134 @@
...
+
+            if force_compute and not metadata_only:
+                channel.data = channel.data.compute()
```
djhoese, line 116: You might want to do `channel = channel.compute()` so that xarray is making sure all dask arrays in `.attrs` or `.coords` are being computed. As written this is only doing the primary data dask array which may not be sufficient.

djhoese: I should probably also mention there are some Satpy edge cases where if you garbage collect the `Scene` and then later try to compute the data you'll get an error because the open file object that the data was read from no longer exists or is no longer open. There are options for handling this, but I'm not sure what's easiest for GeoIPS.

djhoese (commented):
Nice job! This looks really good to me, but I have lots of suggestions. This being a generic reader also screams "...but what would a specific version look like?" or "but how do we list all readers that Satpy can use?". I've been talking with Mindy and Jeremy today at the workshop and lots of brainstorming has been happening so probably not something you need to worry about now.

Read my comments so far and let me know what you think though.

geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,134 @@
...
+# # # https://github.com/NRLMMD-GEOIPS.
+
+
+"""Generic Satpy reader.
+
+Provides a GeoIPS wrapper for Satpy reader functionality. For more detailed
+documentation of Satpy, see the documentation here:
+https://satpy.readthedocs.io/en/stable/. This reader creates a Scene object with
+the provided filenames and specified reader, then loads provided channels. (The
+satpy_reader, and fnames arguments are passed to the Scene constructor)  The
```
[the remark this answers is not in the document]
robertdemariacira: The parentheses portion is its own sentence but is intended to comment on the previous sentence to indicate how these arguments will be used.  This can be rewritten to be more clear.

geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,134 @@
...
+def call(
+    fnames: List[str],
+    source_name: str,
+    platform_name: str,
+    metadata_only: bool = False,
+    chans: Optional[List[str]] = None,
+    area_def: Optional[pyresample.geometry.AreaDefinition] = None,
+    self_register: bool = False,
+    satpy_reader: str = "abi_L1b",
+    channel_groups: Dict[str, List[str]] = {"0.5km": ["C02"]},
```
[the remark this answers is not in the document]
robertdemariacira: Agreed that these should probably be changed to required arguments if possible.  I wrote most of this at a GeoIPS workshop so my lack of GeoIPS knowledge will also be apparent throughout the code here.

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: I don't know if @mindyls or @jsolbrig would disagree, but my thought is that this plugin should probably focus on handling simple use cases.  More advanced use cases, such as loading composites, should probably have their own reader plugin.  So my preference would be to simply remove the `generate` kwarg.

djhoese: To add more context to this, the `generate` kwarg is typically used as an optimization by users who create a lot of RGB composites from only a few input bands. By specifying `generate=False` you could resample, for example, 4 Level 1 bands but produce many RGBs from those 4 inputs. The default, if no resampling/interpolation was needed, would be to resample each of the RGB bands individually assuming they were different from each other when they might not be (ex. R in composite 1 is the same as G in composite 2, but Satpy doesn't know that).

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: At the moment, I think everything downstream from the reader plugin assumes that it's receiving numpy arrays. (Please correct me if I'm wrong @mindyls and @jsolbrig). So to me it makes sense to make sure that `get_lonlats()` is generating a dask array that is then forced to generate numpy arrays.

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: Good point.  I forgot that the metadata may also have dask arrays.

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: I think the implementation you described makes sense.  It would be nice to allow people to load multiple calibrations for the same channel.

geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,145 @@
...
+
+DEFAULT_COORD_CHUNK = 4096
+
+
+def call(
+    fnames: List[str],
+    source_name: str,
+    satpy_reader: str,
+    channel_groups: Dict[str, List[str]],
+    platform_name: Optional[str] = None,
```
djhoese, line 59: Why is `platform_name` needed? Is GeoIPS really strict on what these can be? Satpy's returned DataArrays should have a platform name in the `.attrs` metadata.

geoips/plugins/modules/readers/generic_satpy.py
djhoese: Yes, you're right. There is the Satpy functionality of a `DataQuery` where instead of string names you can pass `DataQuery(name="C01", calibration="radiance")` and other combinations of things that show up in a DataID.

geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,145 @@
...
+        is_first_channel = True
+        data_dict = {}
+        for channel_name in channel_names:
+            channel = scene[channel_name]
+            if is_first_channel:
+                is_first_channel = False
+                lons, lats = _compute_coords(channel, coord_chunks)
+
+            if force_compute and not metadata_only:
+                channel = channel.compute()
```
djhoese, line 113: Your comment about making sure that lons and lats are computed at the same time made me wonder if this channel compute should be somehow computed later so that all datasets/products are computed at the same time. In most cases for datasets coming straight from a reader's file this won't matter. However, in cases where there are shared dask computations then doing the compute all together could be a big deal. I *think* you can pass any python native data structure to `dask.array.compute` and it will recurse through it to find dask arrays to compute. So you could take `output_datasets` and do `return dask.array.compute(output_datasets)`. Again, I think, it should be tested.

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: I think at the time I started this, it was required by GeoIPS, but it doesn't look like it is now.  If @mindyls or @jsolbrig can confirm that the `platform_name` isn't a required attribute in the output Dataset then I think it makes sense to remove it.

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: Finally getting back to this.  I was originally thinking of altering the code to support allowing multiple calibrations.  However, after evaluating how much time I have to work on this plugin as well as going with the philosophy that this plugin would just address basic usage of Satpy, I'll just go with your initial recommendation of removing the filter kwargs.

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: I'll give this a try.  If this doesn't work, then it may be possible to write some simple alternative logic that would accomplish the same thing.

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: It looks like the unit tests require the `platform_name` to be present in each output `Dataset`.

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: For the time being, I'm going to leave this one as a possible future improvement/optimization.

geoips/plugins/modules/readers/generic_satpy.py
mindyls: Correct - at the moment dask arrays passed through GeoIPS would probably be a negative (because they'd end up being computed a bunch of times during sectoring / interpolating).  Computing them once in the reader at this point probably makes sense.  We'd have to implement better support for dask arrays throughout the standard geoips tools.

geoips/plugins/modules/readers/generic_satpy.py
mindyls: platform_name is a required attribute on our xarray dataset outputs from the readers: https://github.com/NRLMMD-GEOIPS/geoips/blob/main/docs/source/devguide/xarray_standards.rst 

It is not strictly enforced what these values are, like it is for source_names, though having consistent platform names throughout GeoIPS would probably good...  So we'd probably want a mapping somewhere of satpy platform names to geoips platform names.

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: At the moment, I'm going to leave this one as a future optimization/improvement.

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: Right now the plugin will look up the `platform_name` from the attributes of the first channel it encounters.  By default, it will use the `platform_name` key but this can be changed using the command line arguments. Additionally, you can optionally just specify the platform name via the command line arguments if necessary.

geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,228 @@
+# # # This source code is protected under the license referenced at
+# # # https://github.com/NRLMMD-GEOIPS.
+
+
+"""Generic Satpy reader.
+
+Provides a GeoIPS wrapper for basi Satpy reader functionality. For more detailed
```
AJThorpe, line 7: Spelling:

GeoIPS wrapper for _basi_ Satpy reader functionality

geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,228 @@
...
+    source_name: str,
+    satpy_reader: str,
+    channel_groups: Dict[str, List[str]],
+    platform_name: Optional[str] = None,
+    platform_name_attr: str = "platform_name",
+    metadata_only: bool = False,
+    chans: Optional[List[str]] = None,
+    area_def: Optional[pyresample.geometry.AreaDefinition] = None,
+    self_register: bool = False,
+    roi: float = 3000,
```
AJThorpe, line 67: Would ROI need to change per instrument? For example if I input high resolution VIIRS data versus low resolution geostationary data?

geoips/plugins/modules/readers/generic_satpy.py
```diff
@@ -0,0 +1,228 @@
...
+        is_first_channel = True
+        data_dict = {}
+        for channel_name in channel_names:
+            channel = scene[channel_name]
+            if is_first_channel:
+                is_first_channel = False
+                lons, lats = _compute_coords(channel, coord_chunks)
+
+                if platform_name is None:
+                    platform_name = channel.attrs[platform_name_attr]
```
AJThorpe, line 172: As per Mindys point, do the platform names and source names match or atleast map to the geoips ones? Products are reliant on the source name and would result in an error if its not mapped correctly.

AJThorpe (commented):
Could you attach a test script that shows off this reader's functionality? Then we could also test it with other data sources, check latency, and see if any fixes need to be made.

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: Fixed typo.

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: At the moment, you can provide the `platform_name` on the command line and fall back on looking it up from the satpy metadata.  If a mapping between the satpy `platform_name` and the geoips `platform_name` needs to occur, would you be able to provide a list of valid geoips `platform_name`s?

Maintaining this mapping may be difficult in the long run. Instead of falling back on the lookup, the code could simply require that the `platform_name` is provided. Would this be an acceptable solution?

geoips/plugins/modules/readers/generic_satpy.py
robertdemariacira: That's a good point. I could make the code ignore the `roi` argument and instead add a dictionary argument that maps channel group names to its corresponding ROI.

robertdemariacira:
Regarding the test script, the code currently provides `get_test_files` and `get_test_parameters` so that it can be included in the geoips unit tests.  Is this sufficient or would it be good to add a shell script in the `tests/scripts/` directory?

If a shell script is required, is there any way to just call the reader plugin? I'm not sure, but I don't believe any existing "downstream" plugins can use the output that this reader plugin produces without crashing.

biosafetylvl5:
We discussed this in our developers meeting. We _should_ be able to use an `imagery_clean` output formatter in a shell script (in `tests/scripts`. @evrose54 is going to put a test script together using your provided testing functions so we can verify this works with two/more data sources. 

There's been some discussion about this going in a dedicated plugin repo, but we're split on that. So it's still good to merge into core GeoIPS 🙂 And test scripts in proxyviz would be awesome as well. 

We really appreciate how well documented this is! Thanks for putting this PR in 🌎 😊 🌎

robertdemariacira:
Thanks!  Yeah, the [ProxyVis plugin](https://github.com/robertdemariacira/geoips_proxy_vis) assumes this plugin is available in the environment and has a set of test scripts that verify that ProxyVis works end-to-end.  So those can help act as a reference for making test scripts for this plugin.

biosafetylvl5:
@AJThorpe is going to comment here and provide a test that allows this to use built-in GeoIPS plugins for a dedicated test. Potentially we could move this to ProxyViz if this approach does not work.

AJThorpe:
@djhoese Quick satpy related question: 
Is there an easy way to load a channel with different calibrations (ex: BT and radiance) without resetting the Scene?

i.e. Currently to load different calibrations I do (ABI data, satpy version 0.58.0):
```
In [6]: scn = satpy.Scene(reader='abi_l1b',filenames=nfiles[:1])
In [7]: scn.available_dataset_ids()
Out[7]:
[DataID(name='C01', wavelength=WavelengthRange(min=0.45, central=0.47, max=0.49, unit='µm'), resolution=1000, calibration=<1>, modifiers=()),
 DataID(name='C01', wavelength=WavelengthRange(min=0.45, central=0.47, max=0.49, unit='µm'), resolution=1000, calibration=<3>, modifiers=()),
 DataID(name='C01', wavelength=WavelengthRange(min=0.45, central=0.47, max=0.49, unit='µm'), resolution=1000, calibration=<5>, modifiers=())]
In [8]: dlist = scn.available_dataset_ids()
In [9]: scn.load([dlist[0]])
In [10]: scn['C01']
Out[10]:
<xarray.DataArray 'Rad' (y: 10848, x: 10848)> Size: 471MB
dask.array<mul, shape=(10848, 10848), dtype=float32, chunksize=(2712, 2712), chunktype=numpy.ndarray>
Coordinates:
    crs      object 8B PROJCRS["unknown",BASEGEOGCRS["unknown",DATUM["Unknown...
  * y        (y) float64 87kB 5.434e+06 5.433e+06 ... -5.433e+06 -5.434e+06
  * x        (x) float64 87kB -5.434e+06 -5.433e+06 ... 5.433e+06 5.434e+06
Attributes: (12/29)
    orbital_parameters:     {'projection_longitude': -75.0, 'projection_latit...
    long_name:              Bidirectional Reflectance
    standard_name:          toa_bidirectional_reflectance
....
In [15]: scn = satpy.Scene(reader='abi_l1b',filenames=nfiles[:1])
In [16]: scn.load([dlist[1]])
In [17]: scn['C01']
Out[17]:
<xarray.DataArray 'Rad' (y: 10848, x: 10848)> Size: 471MB
dask.array<add, shape=(10848, 10848), dtype=float32, chunksize=(2712, 2712), chunktype=numpy.ndarray>
Coordinates:
    crs      object 8B PROJCRS["unknown",BASEGEOGCRS["unknown",DATUM["Unknown...
  * y        (y) float64 87kB 5.434e+06 5.433e+06 ... -5.433e+06 -5.434e+06
  * x        (x) float64 87kB -5.434e+06 -5.433e+06 ... 5.433e+06 5.434e+06
Attributes: (12/29)
    orbital_parameters:     {'projection_longitude': -75.0, 'projection_latit...
    long_name:              ABI L1b Radiances
    standard_name:          toa_outgoing_radiance_per_unit_wavelength
....
```

djhoese:
So the key to your problem is the `DataQuery` object (`from satpy import DataQuery` I think should work). You should be able to load, theoretically, multiple calibrations by specifying both DataIDs to the Scene's `load` call. They both live in the Scene but only one is accessed by the generic string name "C01". You would want to use `DataQuery(name="C01", calibration="radiance")` to access the other. You could make a DataQuery object for the BT/Reflectance calibration too.

biosafetylvl5:
Given SatPys new license, it looks like we can go ahead and merge this into main if we'd like!

PR #732, opened 2024-08-07
Adding NOAA-20, NOAA-21, S-NPP AWS downloads

mindyls:
Required to fix GEOIPS/geoips#666

This is required to download NOAA VIIRS test datasets for the data fusion tutorial.

```
Package: geoips_full
Total run time: 2017 seconds
Number data types run: 56
Number data types failed: 0
Wed Aug  7 21:34:26 UTC 2024
```


christopher-selman (approved):

tests/download_noaa_aws.sh
```diff
@@ -87,7 +104,6 @@ elif [[ "$satellite" == "geokompsat" ]]; then
     echo "rclone --config $rclone_conf lsf $rclone_path"
     files=`rclone --config $rclone_conf lsf $rclone_path`
     echo "COMPARE: ${yyyy}${mm}${dd}${hh}${mn}"
-    break
```
mindyls, line 90: I don't think this was supposed to be here under the geokompsat section

cpcamacho: I do not think so either! Resolving this conversation

evrose54 (approved):
LGTM; not sure if this is showing up for any of you but we might want to look at this before merging. Screenshot 2024-08-07 at 3 43 17 PM

tests/download_noaa_aws.sh
```diff
@@ -25,13 +25,30 @@ if [[ "$8" == "" && -z "$GEOIPS_PACKAGES_DIR" ]]; then
...
     echo "       \$GEOIPS_TESTDATA_DIR/test_data_noaa_aws/data/<satellite>/<YYYYmmdd>/<HHMN>"
-    echo "    rclone_conf: if 'default' or not specified, defaults to:"
+    echo "    rclone_conf: "
+    echo "        if 'default' or not specified, defaults to:"
     echo "        \$GEOIPS_PACKAGES_DIR/geoips/setup/rclone_setup/rclone.conf"
-    echo "    wildcard_list: list of strings to match in filenames (ie, channels). Defaults to all files for dtg"
+    echo "    collection:"
+    echo "        viirs"
+    echo "           VIIRS-IMG-GEO-TC"
+    echo "           VIIRS-I5-SDR"
```
mindyls, line 48: How do we differentiate between "collection' and "wildcard_list" ?  Is collection used for all of the satellites?

"collection' is the subdirectory, and "wildcard_list" are substrings from the filenames?  Is this always true?

AJThorpe (approved):

christopher-selman (approved):

tests/download_noaa_aws.sh
```diff
@@ -25,13 +25,30 @@ if [[ "$8" == "" && -z "$GEOIPS_PACKAGES_DIR" ]]; then
...
+    echo "        himawari8"
+    echo "        himawari9"
+    echo "        geokompsat"
+    echo "        noaa-20"
+    echo "        noaa-21"
+    echo "        snpp"
+    echo "        jpss"
+    echo "    testdata_dir: "
+    echo "        if 'default' or not specified, defaults to: "
     echo "       \$GEOIPS_TESTDATA_DIR/test_data_noaa_aws/data/<satellite>/<YYYYmmdd>/<HHMN>"
```
jsolbrig, line 41: I wonder if, in the future, this should default to `test_data_noaa_<satellite_name>`.

jsolbrig (dismissed):

tests/download_noaa_aws.sh
cpcamacho: I'm not entirely sure if "collection" is the right terminology, but yes, that defines the subdirectory under the S3 bucket. "wildcard_list" only applies to the substrings in the filenames. I think that will always be true, but I could be wrong.

tests/download_noaa_aws.sh
mindyls: We'll consider that whne we switch to Evan's python downloader.

cpcamacho (approved):

jsolbrig (approved):

PR #735, opened 2024-08-08
Refactor base paths and fix make_dirs bug

biosafetylvl5:
# Reviewer Checklist

* [x] Required ***existing tests*** pass 
* [x] NO REQUIRED ***unit tests*** (no new functionality)
* [x] NO REQUIRED ***integration tests*** (no new functionality)
* [x] Required ***documentation*** updated
* [x] NO REQUIRED ***updates to other repos*** (no change to functionality)

# Related Issues
## Original file very difficult to read and comprehend. Difficult to maintain.

That's it.

## Found Issue
Fixes found issue with make_dirs. Previously, if make_dirs was passed a read only directory (eg. `/read_only`), it would print:

```
                "/read_only: We thought /read_only did not exist, but then it did. "
                "Not trying to make directory",
```

This is also true for anything that could trigger any of these OS Errors:

```
├── OSError
      │    ├── BlockingIOError
      │    ├── ChildProcessError
      │    ├── ConnectionError
      │    │    ├── BrokenPipeError
      │    │    ├── ConnectionAbortedError
      │    │    ├── ConnectionRefusedError
      │    │    └── ConnectionResetError
      │    ├── FileExistsError
      │    ├── FileNotFoundError
      │    ├── InterruptedError
      │    ├── IsADirectoryError
      │    ├── NotADirectoryError
      │    ├── PermissionError
      │    ├── ProcessLookupError
      │    └── TimeoutError
```

I changed it to use the built in `exist_ok` flag to fix this behaviour.

# Testing Instructions
0. Use this test file in `geoips/filenames/` after saving the `base_paths.py` file from main into `base_paths_old.py` (updated by @jsolbrig; the old version is in the "OLD VERSION" section below):
```python
from geoips.filenames import base_paths as base_paths_new
from geoips.filenames import base_paths_old
import os

new_paths = base_paths_new.PATHS
old_paths = base_paths_old.PATHS

compare_values = []

for key in old_paths:
    same = False
    print(f"VARIABLE: {key}")
    if old_paths[key] == new_paths[key]:
        print("\tValues are identical.")
        same = True
    elif os.path.samefile(old_paths[key], new_paths[key]):
        print("\tPaths point to the same file.")
        same = True
    if not same:
        try:
            if os.path.abspath(old_paths[key]) == os.path.abspath(new_paths[key]):
                same = True
                print("\tAbsolute paths match.")
        except Exception:
            pass
    if not same:
        print("\tERROR VALUES ARE DIFFERENT")
        print("\tOld Value", old_paths[key])
        print("\tNew Value", new_paths[key])
    compare_values.append(same)

print("\n")
if sum(compare_values) == len(compare_values):
    print("All values same.")
else:
    print("ERROR, VALUES ARE NOT SAME.")
```

## OLD VERSION
  ```python
  import base_paths as base_paths_new
  import base_paths_old
  import os

  new_paths = base_paths_new.PATHS
  old_paths = base_paths_old.PATHS

  compare_values = []

  for key in old_paths:
      same = False
      print(f"VARIABLE: {key}")
      if old_paths[key] == new_paths[key]:
          print("\tValues are identical.")
          same = True
      if not old_paths[key] == new_paths[key]:
          try:
              if os.path.abspath(old_paths[key]) == os.path.abspath(new_paths[key]):
                  same = True
                  print("\tAbsolute paths match.")
          except Exception:
              pass
      if same == False:
          print("\tERROR VALUES ARE DIFFERENT")
          print("\tOld Value", old_paths[key])
          print("\tNew Value", new_paths[key])
      compare_values.append(same)

  print("\n")
  if sum(compare_values) == len(compare_values):
      print("All values same.")
  else:
      print("ERROR, VALUES ARE NOT SAME.")
  ```

2. Run base test
3. Run Full test
4. Make sure NRL runs with internal envs variables

# Summary
I refactored `geoips/filenames/base_paths.py` so that it is more readable. I updated all of the docstrings in the file. I fixed a bug with `make_dirs`. 

# Output

test file:
```
geoips_user@d25fe1b3d4d3:/app/geoips_packages/geoips/geoips/filenames$ python3 test.py 
VARIABLE: BASE_PATH
	Absolute paths match.
VARIABLE: GEOIPS_DOCS_URL
	Values are identical.
VARIABLE: GEOIPS_OPERATIONAL_USER
	Values are identical.
VARIABLE: GEOIPS_OUTDIRS
	Values are identical.
VARIABLE: GEOIPS_PACKAGES_DIR
	Values are identical.
VARIABLE: GEOIPS_BASEDIR
	Absolute paths match.
VARIABLE: GEOIPS_TESTDATA_DIR
	Values are identical.
VARIABLE: GEOIPS_DEPENDENCIES_DIR
	Values are identical.
VARIABLE: PRESECTORED_DATA_PATH
	Values are identical.
VARIABLE: PREREAD_DATA_PATH
	Values are identical.
VARIABLE: PREREGISTERED_DATA_PATH
	Values are identical.
VARIABLE: PRECALCULATED_DATA_PATH
	Values are identical.
VARIABLE: CLEAN_IMAGERY_PATH
	Values are identical.
VARIABLE: ANNOTATED_IMAGERY_PATH
	Values are identical.
VARIABLE: FINAL_DATA_PATH
	Values are identical.
VARIABLE: PREGENERATED_GEOLOCATION_PATH
	Values are identical.
VARIABLE: GEOIPS_COPYRIGHT
	Values are identical.
VARIABLE: GEOIPS_COPYRIGHT_ABBREVIATED
	Values are identical.
VARIABLE: GEOIPS_RCFILE
	Values are identical.
VARIABLE: TC_TEMPLATE
	Absolute paths match.
VARIABLE: DEFAULT_QUEUE
	Values are identical.
VARIABLE: BOXNAME
	Values are identical.
VARIABLE: HOME
	Values are identical.
VARIABLE: SCRATCH
	Values are identical.
VARIABLE: LOCALSCRATCH
	Values are identical.
VARIABLE: SHAREDSCRATCH
	Values are identical.
VARIABLE: GEOIPS_ANCILDAT
	Values are identical.
VARIABLE: GEOIPS_ANCILDAT_AUTOGEN
	Values are identical.
VARIABLE: LOGDIR
	Values are identical.
VARIABLE: GEOIPSDATA
	Values are identical.
VARIABLE: GEOIPS_VERS
	Values are identical.
VARIABLE: TCWWW
	Values are identical.
VARIABLE: TCWWW_URL
	Values are identical.
VARIABLE: PUBLICWWW
	Values are identical.
VARIABLE: PUBLICWWW_URL
	Values are identical.
VARIABLE: PRIVATEWWW
	Values are identical.
VARIABLE: PRIVATEWWW_URL
	Values are identical.
VARIABLE: TC_DECKS_DB
	Values are identical.
VARIABLE: TC_DECKS_DIR
	Values are identical.


All values same.
```

Base test:
```
Package: geoips_base
Total run time: 80 seconds
Number data types run: 3
Number data types failed: 0
Thu Aug  8 21:05:17 UTC 2024
```


jsolbrig:
I updated the testing instructions with a new test script. It accounts for symlinks by using `os.path.samefile`. This could cause problems if one of the URLs differs, though, since `os.path.samefile` will error if given URLs to compare.

geoips/filenames/base_paths.py
```diff
@@ -1,252 +1,171 @@
...
-if getenv("TC_DECKS_DB"):
-    PATHS["TC_DECKS_DB"] = getenv("TC_DECKS_DB")
-
-if getenv("TC_DECKS_DIR"):
-    PATHS["TC_DECKS_DIR"] = getenv("TC_DECKS_DIR").rstrip("/")
-else:
-    PATHS["TC_DECKS_DIR"] = pathjoin(
-        PATHS["GEOIPS_OUTDIRS"], "longterm_files", "tc", "decks"
-    )
+    geoips_global_variables = {
```
mindyls, line 94: Maybe leave a comment here that these are the explicitly defined global variables (not dependent on another variable as a parent directory), pulling from os.getenv if it is specified, else use the defined default.

geoips/filenames/base_paths.py
```diff
@@ -1,252 +1,171 @@
...
+        # Configuration and Queue
+        "GEOIPS_RCFILE": os.getenv("GEOIPS_RCFILE", ""),
+        "DEFAULT_QUEUE": os.getenv("DEFAULT_QUEUE", None),
+        # Computer Identifier
+        "BOXNAME": socket.gethostname(),
+    }
+
+    paths.update(geoips_global_variables)
+
+    derivative_directory_paths = {
```
mindyls, line 115: Maybe name this derivative_directory_paths_defaults , and leave a comment that this is specifically the defaults for any path based environment variables,  which all default to locations under either BASE_PATHS or GEOIPS_OUTDIRS, to keep all geoips code and outputs consolidated unless explicitly specified through environment variables.  Also that eventually the environment variable overrides will be replaced (or supplemented) with a yaml or json config.

geoips/filenames/base_paths.py
```diff
@@ -1,252 +1,171 @@
...
+            # Tropical Cyclone Paths
+            "TC_DECKS_DB": "longterm_files/tc/tc_decks.db",
+            "TC_DECKS_DIR": "longterm_files/tc/decks",
+        },
+        paths["BASE_PATH"]: {
+            "TC_TEMPLATE": "plugins/yaml/sectors/dynamic/tc_web_template.yaml",
+        },
+    }
+
+    for top_directory, sub_directories in derivative_directory_paths.items():
```
mindyls, line 153: Leave a comment here that we are looping through all the directory-based paths, and using "get_env_var" function to set the paths to the environment variable specified option when defined, else defaulting to the passed-in default (second argument).  Also note that in the future, this will likely also pull from a YAML or JSON config file (maybe still support env var override?  Not sure).

geoips/filenames/base_paths.py
mindyls: Maybe these should also use "get_env_var" instead of specifying the defaults directly in the dict, for consistency and future-proofing for when we are also pulling from the YAML/JSON config?

geoips/filenames/base_paths.py
```diff
@@ -1,252 +1,171 @@
...
+        paths["HOME"] = os.getenv("HOMEDRIVE") + os.getenv("HOMEPATH")
+    else:
+        paths["HOME"] = os.getenv("HOME").rstrip("/")
+
+    # Setting links for WWW Paths
+    www_paths = ["TCWWW", "PUBLICWWW", "PRIVATEWWW"]
+    for path in www_paths:
+        paths[f"{path}_URL"] = get_env_var(f"{path}_URL", paths[path])
+
+    return paths
```
mindyls, line 168: Do we need to update paths with www_paths?  Looks like only the URL version are in here now?

geoips/filenames/base_paths.py
mindyls: Maybe these should also use "get_env_var" instead of specifying the defaults directly in the dict, for consistency and future-proofing for when we are also pulling from the YAML/JSON config?  If we udpated this, then these should probably be named "geoips_global_variables_defaults", because the actual paths would be set in a loop below calling "get_env_var".

geoips/filenames/base_paths.py
biosafetylvl5: I'm not sure what you mean here

geoips/filenames/base_paths.py
biosafetylvl5: You're right. Using `get_env_var` makes way more sense here. My bad! Will fix.

geoips/filenames/base_paths.py
biosafetylvl5: I update paths directly in this block, which is a bit different than how I did it above.

geoips/filenames/base_paths.py
biosafetylvl5: Added! I put some of this info in the module docstring since it applies to the whole file.

geoips/filenames/base_paths.py
biosafetylvl5: Added! :)

geoips/filenames/base_paths.py
biosafetylvl5: Fixed!

jsolbrig (approved):
This looks good to me! I think it's a good step towards having a main configuration file.

geoips/filenames/base_paths.py
mindyls:
```
    www_paths = ["TCWWW", "PUBLICWWW", "PRIVATEWWW"]
    for path in www_paths:
        paths[f"{path}_URL"] = get_env_var(f"{path}_URL", paths[path])
```
TCWWW, PUBLICWWW, and PRIVATEWWW are NOT in paths dictionary, correct?  They were before.

biosafetylvl5: They ARE in the paths dictionary, at least from what I'm seeing

Above, I set them to:

```
            # WWW Paths
            "TCWWW": "preprocessed/tcwww",
            "PUBLICWWW": "preprocessed/publicwww",
            "PRIVATEWWW": "preprocessed/privatewww",
```

And then in the for loop you shared I set them to the same value (which I _believe_ was what was happening in the og code)

mindyls: Oh, nevermind, I didn't see them up ther.e

geoips/filenames/base_paths.py
mindyls: It should probably be named "default_geoips_global_variables" for consistency?  Because the final values are set below?

mindyls (approved):

jsolbrig (approved):

mindyls (approved):

Issue #684, opened 2024-07-17
Add committing of release notes on merge


biosafetylvl5:

biosafetylvl5:
Done. Thanks @mindyls !

PR #746, opened 2024-08-15
Add brassy build workflow on new tag

biosafetylvl5:
# GitHub Actions Workflow: Generate Release Note with Brassy

## Overview
This GitHub Actions workflow automates the generation and management of release notes. The workflow is triggered either manually via `workflow_dispatch` or automatically when a new tag is pushed.

## Job: Generate Release

1. **Checkout Code**
   - The code is checked out from the repository with all history and tags.

2. **Extract Branch Name**
   - The current branch name is extracted using `git`, and it is stored in the GitHub Actions output.

3. **Extract Version**
   - The latest version tag (excluding certain tags) is extracted from the repos new tag

4. **Install Brassy**

5. **Run Brassy on File**
   - A release note file is generated using `brassy`. A temporary header file is created for the release note, and the output is saved to a file named based on the extracted version.

6. **Commit New Release Note and Delete Old YAML Files**
   - The new release note is committed to a new branch named after the current branch with a prefix `update-release-note-`.
   - Old YAML files in the release notes directory are deleted and committed.

7. **Push New Release Note to Branch**
   - The new branch is pushed to the repository.

8. **Create Pull Request**
   - A pull request is automatically created to merge the new release note branch back into the current branch.
   - Example: https://github.com/NRLMMD-GEOIPS/geoips/pull/749


biosafetylvl5:
Fixes #684

mindyls (approved):
Looks good to me, excited to merge and tag/release and see what happens!

.github/workflows/new-brassy-note.yaml
```diff
@@ -0,0 +1,92 @@
+name: Generate Release Note with Brassy
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
```
mindyls, line 6: Oh, we probably explicitly only want this to run on tag/release, because otherwise the most recent tag will be the previous version, so we will be overwriting (?) the previous versions' release notes.

"update_this_release_note" would tell you the correct release note file to write to in between releases - both for every PR merge, as well as immediately after tag/release.  I tag/release the current version on the repository BEFORE updating "update_this_release_note" - the first thing merged in after a version release is the new "update_this_release_note", so it should always be accurate.

If we run this on every PR merge, then the plugin repositories would have to clone the geoips repo in this CI in order to check what "update_this_release_note" is, to see which release note to create.

cpcamacho (approved):
LGTM

Issue #714, opened 2024-07-26
Discuss how to share LOG attribute in the CLI


evrose54:
# Requested Update

### Description
PR #659 introduced a new shared argument ``--log-level`` which sets the level in which we'll output information with a logger object. This argument is now working, however it's not easy to access the ``LOG`` attribute that will be shared amongst all command classes. Currently, the way to access LOG goes as such:

```python
self.LOG = getattr(LOG, self.GEOIPS_ARGS.log_level)
self.LOG("LOG LEVEL = {self.GEOIPS_ARGS.log_level}")
```
This sets the logger object at the correct log level, but this does not actually share the same logger object with all command classes. We need to figure out a method to parse ``--log-level`` before all command classes are initialized, so they can all inherit the same logger object from ``GeoipsCommand``. 

### Background and Motivation
This comes from this [comment](https://github.com/NRLMMD-GEOIPS/geoips/pull/659#discussion_r1692179153) on PR #659.
  
### Environment
*  [ IF APPROPRIATE: Corresponding repository versions ]

### Code to demonstrate issue
``geoips.commandline.commandline_interface:GeoipsCLI:execute_command``

PR #759, opened 2024-08-26
Share log attr between cli classes

evrose54:
# Reviewer Checklist

# Related Issues
fixes NRLMMD-GEOIPS/geoips#714

# Testing Instructions
``pytest -v /tests/unit_tests/commandline/.``

# Summary
Since we've started developing the GeoIPS CLI, we've known that there was a need to
have a shared logger object among all of the CLI command classes for debugging
purposes and to output basic information to the user if requested. We can now do
this using the '--log-level' / '-log' flag. Depending on the log level set, all
levels that match or are of higher 'order' than the log level provided will be
outputted. This functionality matches that used in 'run_procflow'. Logging is now
a supported utility throughout the CLI, and can be used in any CLI class via
self.LOG.<log_level>(log_statement).

# Output
I didn't add any permanent logging statements but we can add some if wanted. I did test the new functionality and it works as expected!


geoips/commandline/geoips_command.py
```diff
@@ -64,14 +65,14 @@ class ParentParsers:
...
+    geoips_parser = argparse.ArgumentParser()
     geoips_parser.add_argument(
-        "--log_level",
         "-log",
+        "--log-level",
         type=str,
         default="interactive",
-        choices=["debug", "error", "info", "interactive", "warning"],
-        help="The logging level to use for output via the CLI.",
+        choices=["interactive", "debug", "info", "warning", "error"],
```
jsolbrig, line 74: I think this line should be reverted to the previous order.

jsolbrig (approved):

geoips/commandline/geoips_command.py
evrose54: It seems ``interactive`` is set to level 35. Debug is 10, info is 20, warning is 30, error is 40, critical is 50. So the list should actually be in the order of ``[debug, info, warning, interactive, error, critical]``.

evrose54: This has since been addressed.

docs/source/releases/latest/share-log-attr-between-cli-classes.yaml
```diff
@@ -0,0 +1,24 @@
...
+    purposes and to output basic information to the user if requested. We can now do
+    this using the '--log-level' / '-log' flag. Depending on the log level set, all
+    levels that match or are of higher 'order' than the log level provided will be
+    outputted. This functionality matches that used in 'run_procflow'. Logging is now
+    a supported utility throughout the CLI, and can be used in any CLI class via
+    self.LOG.<log_level>(log_statement). Note that self.LOG.<log_level> does not have to
+    match the provided '--log-level' ran in the command. This flag is strictly just the
+    LOG level filter. I.e. if '--log-level info' was provided, levels
+    ['info', 'debug', 'interactive'] would be shown. The lower the level, the more
+    logging channels will output.
```
mindyls, line 16: Is this just to avoid having a duplicate logger set up between the procflows and the CLI ?

mindyls (approved):

mindyls (approved):
I left one comment just for clarification on exactly why this is needed.  Looks good to merge.

docs/source/releases/latest/share-log-attr-between-cli-classes.yaml
jsolbrig: Before this, the CLI wasn't making use of logging at all so we could only `print` from the CLI. This allows using the logger for both the CLI and the downstream procflow rather than just the procflow.

PR #763, opened 2024-08-28
New Product Family unsectored_xarray_dict_to_algorithm_to_output_format

evrose54:
Reviewer Checklist

# Related Issues
No issue for this PR.

# Testing Instructions
Run ``/tests/integration_tests/full_test.sh``

# Summary
This branch supports a new product/product_defaults family called
unsectored_xarray_dict_to_algorithm_to_output_format. While this family was
referenced in single_source.py, there was no actual schema that supported this
family. Therefore, we've created that new schema, and updated parts of single_source
to support using the 'unprojected_image' output formatter for this new family. All
in all, by making use of these changes, we can avoid using the procflow arguments
'self_register_dataset' and 'self_register_source', which would reproject geospatial
datasets onto themselves. This was inefficent, and depending on the size of the
dataset, could add close to another minute of processing to your workflow.

# Output
Here is the output of the new test script ``sever.WV-Upper.no_self_register.unprojected_image.sh``
image


docs/source/releases/latest/new-prd-fam-unsectored-xarray-dict-to-algorithm-to-output-format.yaml
```diff
@@ -0,0 +1,24 @@
+performance:
+- description: |
+    This branch supports a new product/product_defaults family called
```
jsolbrig, line 3: This is a nitpick. It's not that "this branch" supports new functionality. You are submitting a change that supports new functionality. So, this should be written from that perspective. For example, "Add support for a new ...".

Sorry for the nitpick, but it will help make release notes more readable.

evrose54: I'll update this!

geoips/plugins/modules/output_formatters/unprojected_image.py
```diff
@@ -60,22 +63,49 @@ def call(
...
+        ]
+    else:
+        slices = [xarray_obj[product_name].data]
+    for slice_idx, slice_data in enumerate(slices):
+        main_ax.clear()
+        # Needs to be flipped upside-down as incoming data is flipped if unprojected.
+        # This is the case for family
+        # unsectored_xarray_dict_to_algorithm_to_output_format, but I'm not sure if it
+        # applies to all other families that use this output formatter. We should look
+        # into this.
```
evrose54, line 79: We should discuss this before we merge this PR.

docs/source/releases/latest/new-prd-fam-unsectored-xarray-dict-to-algorithm-to-output-format.yaml
evrose54: Done

jsolbrig:
I'm good with merging this as a band-aid for the time being, but would like to figure out how to generalize most of the new functionality here. @evrose54 will write an issue discussing what we talked about and link it here. Once done, I'm good with merging this.

jsolbrig (approved):

evrose54:
@jsolbrig see #801.

geoips/plugins/modules/output_formatters/unprojected_image.py
jsolbrig: Unfortunately, I don't think this always works and we may need to figure out a different solution. Either that or we will need to chase down the reason for this.

When I run one of the existing tests, the result is now flipped and incorrect. I ran `./ami.WV-Upper.unprojected_image.sh` which should produce the first image below and now produces the second image.

old

new

evrose54: You're right that this should probably not flip the data upside down. Output formatters should just output, and not manipulate the data. Let me readdress the reader in my OVERCAST package and see if I can flip the data upside down then (or in my algorithms), instead of at this step.

evrose54: Yeah, I can revert that change. The OVCST data is naturally going from south to north and I can fix that within my reader. We should be able to resolve this once I've got that pushed.

evrose54: Pushed those changes; Jeremy feel free to resolve this if you're good with them!

jsolbrig (approved):

jsolbrig (approved):

mindyls (approved):
This is great, thanks for doing it!

jsolbrig (approved):

Issue #647, opened 2024-06-19
Automating ``create_plugin_registries``


evrose54:
# Requested Update

### Description
After some conversation, we've came up with some ideas on how to automate ``create_plugin_registries`` to ensure the plugin registries are in a correct state whenever GeoIPS is ran. 

To start, we'll need to attach a pre-build script to plugin_package installation process, that generates a text file (or other suitable file type) which includes the installation time of that certain package. If this installation time is newer than the modification time of any plugin package's registered_plugins.json (assuming one exists), then run / rerun ``create_plugin_registries`` to ensure GeoIPS is in a correct / valid state.

I think it'd probably be a good idea to attach this functionality to the top level ``__init__.py`` file of all registered GeoIPS plugin packages. It'd be optimal to have this attached to non-registered plugin packages' top level ``__init__.py`` files, so maybe it'd be a good idea to discuss whether or not we want to add this to [Template Basic Plugin](https://github.com/NRLMMD-GEOIPS/template_basic_plugin) in the future.

### Background and Motivation
This issue comes from the last bullet in this issue https://github.com/NRLMMD-GEOIPS/geoips/issues/541.

### Code to demonstrate issue
``geoips.create_plugin_regstries`` and ``geoips.plugin_registry``.

### Checklist for Completion
* [ ] Create a pre-build script which generates the installation time file for each plugin package.
* [ ] Use the information in this file to determine whether or not we should run / re-run ``create_plugin_registries``

PR #781, opened 2024-09-16
Automate ``create_plugin_registries``

evrose54:
# Reviewer Checklist

* [x] Required ***updates to other repos*** added for new/modified functionality

# Related Issues
fixes NRLMMD-GEOIPS/geoips#647
Correlated [PR](https://github.com/NRLMMD-GEOIPS/data_fusion/pull/36) in data_fusion is required to fix a slowdown to data_fusion procflows which was introduced by this PR.

**We can close the correlated PR if we decide to keep this PR as-is. A discussion needs to occur about the downstream impacts of defaulting, or not defaulting, to rebuilding the plugin registries on a plugin retrieval failure.**

# Testing Instructions
Run ``./tests/integration_tests/full_test.sh``

# Summary
The command 'create_plugin_registries' has greatly improved the time it takes to
operate GeoIPS under the hood. This is because all plugins are no longer loaded into
memory, as we can now traverse a plugin registry to load the exact plugins we need
to, when we need them. Unfortuneately, this requires the user to run
'create_plugin_registries' anytime a new plugin has been added to one or more plugin
packages. We've been thinking for a long time of how to automate this process, and
this update does just that.

'create_plugin_registries' can now either be ran manually, or will attempt to be
done automatically by GeoIPS when a plugin is requested. If a user has no registries
in any of their plugin packages, then this is ran the first time get_plugin is
called when certain attributes of the associated interface are being initialized.

If the registries already exist, but a plugin is missing or seemingly out of sync,
then 'create_plugin_registries' will be ran, and get_plugin will be called again. If
the failure persists, then it is likely a problem on the user's end that they need
to fix before running GeoIPS. Now, users no longer have to run
'create_plugin_registries', unless they choose to do so manually.

File Updates:

    added:
      - tests/unit_tests/plugins/test_get_plugin.py
    modified:
      - docs/source/concepts/functionality/plugin-registries.rst
      - geoips/commandline/args.py
      - geoips/commandline/geoips_run.py
      - geoips/commandline/geoips_test.py
      - geoips/filenames/base_paths.py
      - geoips/interfaces/base.py
      - geoips/interfaces/module_based/output_checkers.py
      - geoips/interfaces/yaml_based/products.py

# Output
```python
from geoips.interfaces import readers

test_plg = readers.get_plugin("fake_plugin")

# OUTPUT

Running create_plugin_registries due to a missing plugin.
16_203926 create_plugin_registries.py:318  INTERACTIVE: Writing /local/home/evan/geoips/geoips_packages/geoips_clavrx/geoips_clavrx/registered_plugins.json
16_203927 create_plugin_registries.py:318  INTERACTIVE: Writing /local/home/evan/geoips/geoips_packages/data_fusion/data_fusion/registered_plugins.json
16_203927 create_plugin_registries.py:318  INTERACTIVE: Writing /local/home/evan/geoips/geoips_packages/recenter_tc/recenter_tc/registered_plugins.json
16_203927 create_plugin_registries.py:318  INTERACTIVE: Writing /local/home/evan/geoips/geoips_packages/geoips_plugin_example/geoips_plugin_example/registered_plugins.json
16_203928 create_plugin_registries.py:820  INTERACTIVE: Skipping module 'interp_scipy' from 'geoips', interface_name is 'None'
16_203928 create_plugin_registries.py:820  INTERACTIVE: Skipping module 'boxdefinitions' from 'geoips', interface_name is 'None'
16_203928 create_plugin_registries.py:820  INTERACTIVE: Skipping module 'interp_pyresample' from 'geoips', interface_name is 'None'
16_203928 create_plugin_registries.py:820  INTERACTIVE: Skipping module 'geostationary_geolocation' from 'geoips', interface_name is 'None'
16_203928 create_plugin_registries.py:820  INTERACTIVE: Skipping module 'hrit_reader' from 'geoips', interface_name is 'None'
16_203928 create_plugin_registries.py:820  INTERACTIVE: Skipping module 'remss_reader' from 'geoips', interface_name is 'None'
16_203928 create_plugin_registries.py:820  INTERACTIVE: Skipping module 'tc_file_naming' from 'geoips', interface_name is 'None'
16_203928 create_plugin_registries.py:318  INTERACTIVE: Writing /local/home/evan/geoips/geoips_packages/geoips/geoips/registered_plugins.json

PluginError: Plugin 'fake_plugin', from interface 'readers' appears to not exist.
Create plugin, then call create_plugin_registries?
```


jsolbrig (changes requested):
This should have a CLI flag added to allow turning on automatic registry creation.

evrose54:
Add ``GEOIPS_REBUILD_REGISTRIES`` to base_paths.py and have this default to false. If set to True, set ``rebulid_registries`` to True for all ``get_plugin`` calls.

jsolbrig:
> have this default to false.

Let's have this default to `True` instead, as discussed with NRL last week.

evrose54:
I have this defaulting to True now, running full test to see if things are working.

geoips/commandline/args.py
```diff
@@ -258,6 +258,18 @@ def add_args(parser, arglist=None, legacy=False):
             help="""Specify logging config level for GeoIPS commands.""",
             type=str.upper,
         )
+    if arglist is None or "rebuild_registries" in arglist:
+        parser.add_argument(
+            "--rebuild_registries",
+            type=bool,
+            choices=[True, False],
+            # Will default to $GEOIPS_REBUILD_REGISTRIES if set, else False
+            default=getenv("GEOIPS_REBUILD_REGISTRIES", False),
```
jsolbrig, line 267: Should this also be `True` rather than `False`?

evrose54: Good catch! I updated this. Thanks Jeremy.

jsolbrig (approved):

geoips/interfaces/base.py
```diff
@@ -319,6 +323,84 @@ def __new__(cls):
...
+        By running 'create_plugin_registries', we automate the registration of plugins
+        in GeoIPS. If the plugin persists not to be found, then we'll raise an
+        appropriate PluginError as denoted by 'err_str'.
+
+        Parameters
+        ----------
+        name: str or tuple(str)
+            - The name of the yaml plugin. Either a single string or a tuple of strings
+              for product plugins.
+        rebuild_registries: boolean (default=False)
```
mindyls, line 377: This does not actually have a default at this point.

geoips/interfaces/base.py
```diff
@@ -288,7 +291,7 @@ class BaseModulePlugin:
     pass
 
 
-class BaseInterface:
+class BaseInterface(abc.ABC):
```
mindyls, line 294: This just allows using the abc.abstractmethod ?

geoips/interfaces/base.py
```diff
@@ -332,8 +414,9 @@ def registered_module_based_plugins(self):
                     self.plugin_registry.registered_plugins["module_based"]
                 )
             except (AttributeError, KeyError):
-                raise PluginRegistryError(
-                    "Plugin registries not found, please run 'create_plugin_registries'"
+                self.call_create_plugin_registries()
+                self._registered_module_based_plugins = (
+                    self.plugin_registry.registered_plugins["module_based"]
```
mindyls, line 419: When would it hit this location of calling create_plugin_registries rather than through get_plugin / retry_get_plugin?

geoips/interfaces/base.py
mindyls: I like having the methods themselves default to False, and the overall environment variable default to True, FWIW.

mindyls (approved):
Just a few comments to review

geoips/interfaces/base.py
evrose54: Could you expand on this a bit? I'm a bit confused by your question. When I wrote this code, I envisioned this ``except`` statement to be hit if the plugin registry itself didn't exist. The commandline uses the plugin registry, and this could also be hit when ``get_plugin`` is ran as well.

tests/unit_tests/plugins/test_get_plugin.py
```diff
@@ -0,0 +1,88 @@
...
+                        if idx > 4:
+                            break
+                        yield (interface_name, (source_name, prod_name), "normal")
+                    break
+            else:
+                for idx, plugin_name in enumerate(
+                    registry[plugin_type][interface_name]
+                ):
+                    # Only add up to the first 5 plugins of this interface
+                    yield (interface_name, plugin_name, "normal")
```
jsolbrig, line 38: I don't see anything here that aligns with the comment, restricting the number of plugins to 5.

tests/unit_tests/plugins/test_get_plugin.py
```diff
@@ -0,0 +1,88 @@
...
+from importlib import reload
+import pytest
+
+from geoips import interfaces
+from geoips.errors import PluginError
+from geoips import plugin_registry as plugin_registry_module
+
+
+def yield_interface_plugin_tuples():
+    """Yield tuples of (interface_name, plugin_name)."""
```
jsolbrig, line 14: Can you please update and improve this docstring so it's easier for someone who is reading this to understand what this is returning and why? It looks like the tuple in the docstring needs another element added and it would be useful to say how the three elements are intended to be used, e.g.:

```
Yield tuples of (interface_name, plugin_name_tuple, expected_outcome) for use in other tests.

Yield tuples to use when attempting to load a plugin in the `test_get_plugin` function. The tuple will contain three elements:

- interface_name: The name of the interface that we want to attempt to load a plugin from.
- plugin_name_tuple: A tuple containing one or two strings defining which plugin should be loaded. Two strings is used for the Products interface and one string is used for all other interfaces.
- expected_output: A string describing the expected outcome. This is used in the `test_get_plugin` function to determine which part of the logic to follow and how to test the results.
```

Note that I'm saying `plugin_name_tuple` because of another comment that would help clean up `test_get_plugin` a bit.

tests/unit_tests/plugins/test_get_plugin.py
```diff
@@ -0,0 +1,88 @@
...
+    # given a valid plugin name
+    if test_result in ["normal", "out_of_sync", "non_existent_path"]:
+        if isinstance(plugin_name, tuple):
+            assert curr_interface.get_plugin(
+                plugin_name[0],
+                plugin_name[1],
+                rebuild_registries=True,
+            )
+        else:
+            assert curr_interface.get_plugin(plugin_name, rebuild_registries=True)
```
jsolbrig, line 84: Another option here to cut down on code would be to always make the plugin name a tuple and do this:
```
assert curr_interface.get_plugin(*plugin_name, rebuild_registries=True)
```

tests/unit_tests/plugins/test_get_plugin.py
```diff
@@ -0,0 +1,88 @@
...
+
+def gen_label(val):
+    """Generate the interface-plugin_name-<test_result> id for pytest."""
+    return f"{val[0]}-{val[1]}-{val[2]}"
+
+
+@pytest.mark.parametrize(
+    "int_plug_tuple", yield_interface_plugin_tuples(), ids=gen_label
+)
+def test_get_plugin(int_plug_tuple):
```
jsolbrig, line 68: I think it would be worth also testing with `rebuild_registries=False` to ensure that we get correct behavior there, too.

tests/unit_tests/plugins/test_get_plugin.py
evrose54: Only reason I didn't do that initially is that when you unpack a string that isn't a tuple, you get something like this:
```python
In [1]: c = "an_example_string"

In [2]: d = ("a_tuple", "of_strings")

In [3]: print(*c)
a n _ e x a m p l e _ s t r i n g

In [4]: print(*d)
a_tuple of_strings
```
I could however always return a tuple, even if it only has one string, and this would then work.

tests/unit_tests/plugins/test_get_plugin.py
evrose54: Agreed. I'll add that test now.

tests/unit_tests/plugins/test_get_plugin.py
evrose54: I'll update this docstring. Are you ok if I use what you provided? It does a good job clarifying what this function does.

tests/unit_tests/plugins/test_get_plugin.py
evrose54: Good catch. I added a conditional to make that comment correct.

tests/unit_tests/plugins/test_get_plugin.py
jsolbrig: Yeah, I think that would make this cleaner if `plugin_name` was always a tuple. Much easier to explain in the docstring, too.

evrose54: Made all of the names tuples so that can be simplified.

tests/unit_tests/plugins/test_get_plugin.py
evrose54: Done!

tests/unit_tests/plugins/test_get_plugin.py
evrose54: Updated using what you posted above.

jsolbrig (approved):

geoips/interfaces/base.py
mindyls: I guess when is get module based plugins called?

mindyls (approved):
Looks good, you can resolve conversations

geoips/interfaces/base.py
evrose54: I've got an answer now. This would be hit if no registries existed, or, somehow, no module plugins existed. Same for yaml plugins in the property below. This is nice as we can automatically run ``create_plugin_registries`` the first time an interface is used, if no registries exist. This would likely be hit the first time a user installs GeoIPS but no plugin registries have been created yet.

mindyls (approved):

docs/source/concepts/functionality/plugin-registries.rst
jsolbrig, line 1: It looks like there is some leftover stuff from the previous documentation:

```


Note

We hope to automate the process of creating/updating the plugin registries in the future to avoid requiring the user to directly call create_plugin_registries.
```

It would be good to include information about running `create_plugin_registries` manually, but I think this part looks out of date.

evrose54: I removed this old comment. Lines 24+ should accomplish how to manually run ``create_plugin_registries``, let me know if you feel otherwise!

jsolbrig (approved):

docs/source/concepts/functionality/plugin-registries.rst
jsolbrig: Thanks @evrose54

PR #839, opened 2024-12-12
Fix insecure download of tar archives in test dataset downloader

biosafetylvl5:
* [ ] NO REQUIRED ***existing tests*** (explain why not required)
* [ ] NO REQUIRED ***unit tests*** (explain why not required)
* [ ] NO REQUIRED ***documentation*** (explain why not required)

# Testing Instructions
Download a test data set and make sure it works.

# Summary
- `--test-data-available` requires an argument. It now errors if none is provided instead of saying `None is NOT available for direct download.`
- Tar files are now filtered to make sure they are extracted into the target directory and are not doing any insecure absolute path extractions.


mindyls (approved):
Thanks!

jsolbrig (approved):

PR #840, opened 2024-12-13
Merge dev staging into main

biosafetylvl5:
Should be fine to merge to `main` if CI is passing because all the PRs into `dev-staging` were approved as if they were merged to `main`.

DO NOT MERGE UNTIL RELEASE PROCESS HAS COMPLETED


docs/source/tutorials/setting-up-for-plugin-development.rst
```diff
@@ -0,0 +1,90 @@
...
+   previous install state, which doesn't include any new updates to this file.
+
+   ::
+
+      pip install -e $MY_PKG_DIR
+
+   The ``-e`` portion of the command above means 'editable', so we can edit the package
+   after it is installed and changes will be reflected in the installed package. Again,
+   the only time you must reinstall is when you edit ``pyproject.toml``, which
+   generally only occurs when you create new module based plugins, and must add them as
```
mindyls, line 71: This is actually out of date, but let's not update here

mindyls (approved):
Only one comment! But I don't think that should be updated on this branch, let's just get this merged

docs/source/tutorials/setting-up-for-plugin-development.rst
biosafetylvl5: It always needs be installed with `-e`, right?

jsolbrig (approved):

docs/source/tutorials/setting-up-for-plugin-development.rst
jsolbrig: It should work without `-e` just fine so long as you're not editing anything. I install without `-e` to test every time.

PR #863, opened 2025-01-09
Add build stable image on tag workflow

mindyls:
This will call a geoips_ci reusable workflow, which is a duplicate of the docker image part of doc-test.yaml.  Should have doc-test.yaml call the reusable-build-docker-image.yaml as well..

Will test via this PR

### DIFF
--- a/.github/workflows/build-stable-image.yaml
+++ b/.github/workflows/build-stable-image.yaml
@@ -1,0 +1,41 @@
+name: Build stable docker image
+
+# 1. Tag and Release - on merge of v*-version-release
+#    * Tag current version
+#    * Release just tagged version
+# 2. Package and Publish - on published release (from #1)
+#    * Build wheel
+#    * Publish to pypi
+# 3. Deploy docs - on published release (from #1)
+#    * pip install geoips
+#    * pip install plugin repo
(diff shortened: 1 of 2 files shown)


.github/workflows/build-stable-image.yaml
```diff
@@ -0,0 +1,41 @@
...
+  workflow_dispatch:
+
+jobs:
+  build-stable-image:
+    name: Build stable docker image
+    # You do not appear to be able to use variables in the "uses" field.
+    # Note since this is run after merge to main, reusable-build-docker-image will
+    # build the stable image specifically, which will subsequently be used by
+    # the plugin packages.
+    uses: NRLMMD-GEOIPS/geoips_ci/.github/workflows/reusable-build-docker-image.yaml@build_stable_image_on_tag
```
mindyls, line 37: I am thinking we want to just merge this as-is, then it will allow us to test the updates in geoips_ci after this has been merged, and tagged/released.  Note we will have to come back here after the fact and update build_stable_image_on_tag to main, once everything is working and merged to main on geoips_ci

mindyls:
Known existing pytest unit test errors in doctest.  These have been failing for all merged PRs, need to resolve in a separate PR.
image

PR #926, opened 2025-03-26
Support creating interfaces in new plugin packages

evrose54:
# Reviewer Checklist

* [x] Required ***unit tests*** modified for new functionality
* [x] NO REQUIRED ***documentation*** (explain why not required) (Will do in #985)

# Related Issues
#985
#981

# Testing Instructions
``pytest -m "base or full"``

# Summary
Will complete later. Basically, this PR adds support for users to create interfaces in other plugin packages. However, if a user is going to do this, they should not implement any plugins that fall under supported *GeoIPS* interfaces. The reason I've created this PR is that I'm developing a NRT driving package that makes use of GeoIPS software. This package includes new plugin types that don't need to exist in GeoIPS. 

I've modified the pyproject.toml of the aforementioned package and modified some GeoIPS code to support plugin registries and interfaces that exist in different namespaces. Essentially, GeoIPS and any package with new interfaces can interact with each other, but don't interfere with their own namespaces.

# Output
Will fill once complete.


geoips/geoips_utils.py
```diff
@@ -22,6 +22,19 @@
...
+    Parameters
+    ----------
+    namespace: str
+        - The namespace containg the requested interface module. I.e.
+          'geoips.plugin_packages'.
+    """
+    package_name = namespace.split(".")[0]
+    return import_module(package_name).interfaces
+
+
```
jsolbrig, line 37: Comment this well to indicate that this only collects interfaces from the package whose name matches the name of the first element of the namespace. So, for `geoips.plugin_packages`, only the `geoips` package could create new interfaces. Likewise, for `geoips_driver.plugin_packages`, only the `geoips_driver` package could create new interfaces.

This needs to also be well-documented in the developer documentation to describe the fact that the first element of the namespace and the core package (i.e. the one that implements interfaces) for that namespace must match. `geoips==geoips`, `geoips_driver==geoips_driver`, etc.

geoips/pydantic/bases.py
```diff
@@ -172,6 +177,8 @@ class PluginModel(FrozenModel):
     for more information about how this is used.
     """
 
+    namespace: ClassVar = "geoips.plugin_packages"
```
jsolbrig, line 180: I want to consider moving this onto apiVersion where we would use something like this:

```
apiVersion: ClassVar = "geoips/v1"
```

This would then require that the core package name for a particular namespace matches the apiVersion namespace as well as the plugin namespace from pyproject.toml. This would keep us moving towards these plugins being able to be used in a cloud-native way.

Keep in mind that we are already going to implement apiVersion. This is just an update to how it is used.

geoips/geoips_utils.py
evrose54: Updated this docstring to reflect that. I'll add an issue that states we need to write documentation in the developers section for this.

geoips/pydantic/bases.py
evrose54: I'm going to wait on this until we know whether or not @Srikant-Kumar-Gampa has already implemented this. Once we know that, I'll either keep it or update this to 'apiVersion'.

srikanth-kumar: This hasn't been implemented in any of my current PRs. I was planning to discuss `apiVersion`, as it’s the natural next step in workflow enhancement after adding `kind` and modifying the workflow structure.

srikanth-kumar: @evrose54 Consider this as not implemented and go ahead with it after discussing with @jsolbrig  in the interest of moving forward with this PR.

geoips/geoips_utils.py
jsolbrig: @evrose54 Did you create this issue? If so, can you please comment here with the issue so it's linked?

geoips/pydantic/bases.py
jsolbrig: @evrose54 Let's go ahead and do this. For now, let's hard-code the apiVersion to "geoips/v1" at the top of `PluginModel`. @srikanth-kumar will handle the more complicated task of making this handle multiple apiVersion strings.

geoips/geoips_utils.py
evrose54: Here it is: #985.

geoips/pydantic/bases.py
evrose54: This should be addressed now. If you wanted to provide a different ``apiVersion`` for a given ``<type>PluginModel``, all you'd have to do is ``apiVersion: ClassVar[str] = '<api_version>'``.

jsolbrig: I think that I am going to say this is fine for now. It works and I have a solution for the next step which is to make use of `apiVersion` to load the appropriate models.

docs/source/releases/latest/south-china-sea-sector.yaml
```diff
@@ -0,0 +1,34 @@
+enhancement:
```
mindyls, line 1: Why is this release note / branch named "south china sea sector" ?  Should the release note have a more informative name?

jsolbrig (approved):

docs/source/releases/latest/south-china-sea-sector.yaml
jsolbrig: I renamed this.

mindyls (approved):

PR #963, opened 2025-04-22
Update dockerfile to multi-stage

biosafetylvl5:
# Reviewer Checklist

**Remove lines that are not applicable, explain if you select "NO REQUIRED"**

* [ ] NO REQUIRED ***existing tests*** (explain why not required)
* [ ] Required ***unit tests*** added and pass for new/modified functionality
* [ ] NO REQUIRED ***unit tests*** (explain why not required)
* [ ] NO REQUIRED ***documentation*** (explain why not required)
* [ ] NO REQUIRED ***release notes*** (explain why not required)
* [ ] Required ***updates to other repos*** complete

# Related Issues
fixes NRLMMD-GEOIPS/geoips#NNN

# Testing Instructions

# Summary

# Output

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,44 +1,201 @@
+###############################################################################
+#                              BASE STAGE
+###############################################################################
 FROM python:3.10-slim-bullseye AS base
 
-RUN apt-get update && apt-get -y upgrade \
-    && apt-get install -y wget git libopenblas-dev \
-    && apt-get clean \
+# Avoid interactive prompts
+ARG DEBIAN_FRONTEND=noninteractive
+
+# Install system dependencies
+RUN apt-get update \
+    && apt-get upgrade -y \
+    && apt-get install -y --no-install-recommends \
+       wget git libopenblas-dev g++ make gfortran \
     && rm -rf /var/lib/apt/lists/*
 
-RUN pip install --no-cache-dir -U pip
+# Upgrade pip
+RUN python -m pip install --no-cache-dir --upgrade pip
 
-FROM base AS doclinttest
+###############################################################################
+#                              BUILD STAGE
+###############################################################################
+FROM base AS build
 
-ARG GEOIPS_PACKAGES_DIR=/packages
+# Build arguments
+ARG USER=geoips_user
+ARG USER_ID=1000
+ARG GROUP_ID=1000
+ARG GEOIPS_BASE_CLONE_DIR=.
 
-WORKDIR $GEOIPS_PACKAGES_DIR
-
+ARG GEOIPS_PACKAGES_DIR=/app/geoips_packages
 ARG GEOIPS_OUTDIRS=/output
-ENV GEOIPS_OUTDIRS=${GEOIPS_OUTDIRS}
-
+ARG GEOIPS_DEPENDENCIES_DIR=/app/dependencies
+ARG GEOIPS_TESTDATA_DIR=/geoips_testdata
 ARG GEOIPS_REPO_URL=https://github.com/NRLMMD-GEOIPS/
-ARG GEOIPS_OUTDIRS=/output
-ARG GEOIPS_DEPENDENCIES_DIR=/mnt/dependencies
-ARG GEOIPS_TESTDATA_DIR=/mnt/geoips_testdata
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,44 +1,201 @@
...
+
+# Now copy all of GeoIPS
+COPY --chown=${USER}:${GROUP_ID} ${GEOIPS_BASE_CLONE_DIR} ./
+
+# Configure Git (avoid "detected dubious ownership" warnings when mounted)
+RUN git config --global --add safe.directory '*'
+
+# Install GeoIPS in editable mode
+RUN python -m pip install --no-cache-dir -e "." \
+    && pip uninstall -y matplotlib && pip install matplotlib==3.9.3 \
```
mindyls, line 88: What's the matplotlib pin here for?

biosafetylvl5: Oh gosh, thanks! Thats old. Removing

biosafetylvl5: Removed!

jsolbrig (changes requested):
See my other comment regarding testing difficulties.

We should also consider how to remove the test data from the images. We've talked about this some in terms of using build-time mounts. The current images are 45GB. I'm fine with that in this PR, but I'd like to create an issue to handle this.

Dockerfile
```diff
@@ -1,44 +1,200 @@
...
+###############################################################################
+FROM build AS system_build
+
+# Install all plugins
+RUN python -m pip install --no-cache-dir -e "$GEOIPS_PACKAGES_DIR/geoips/" \
+    && bash $GEOIPS_PACKAGES_DIR/geoips/tests/integration_tests/base_install.sh \
+    && bash $GEOIPS_PACKAGES_DIR/geoips/tests/integration_tests/full_install.sh \
+    && bash $GEOIPS_PACKAGES_DIR/geoips/tests/integration_tests/system_install.sh \
+    && create_plugin_registries
+
```
jsolbrig, line 112: Should this stage build off of the `full_build` stage and just call `bash $GEOIPS_PACKAGES_DIR/geoips/tests/integration_tests/system_install.sh; create_plugin_registries`

Dockerfile
```diff
@@ -1,44 +1,200 @@
...
+ENTRYPOINT ["pytest"]
+
+###############################################################################
+#                           DEV STAGE
+###############################################################################
+FROM test_system AS dev
+
+# Install lint, debug, etc. on top of full test
+RUN python -m pip install --no-cache-dir -e "$GEOIPS_PACKAGES_DIR/geoips/[doc,test,lint,debug]"
```
jsolbrig, line 152: I think we need to reset the ENTRYPOINT here to `/bin/bash` or just unset it.

Dockerfile
```diff
@@ -1,44 +1,200 @@
...
+    && apt-get autoremove -y \
+    && rm -rf /var/lib/apt/lists/*
+
+# Switch back to non-root for runtime
+USER ${USER}
+WORKDIR ${GEOIPS_PACKAGES_DIR}/geoips
+
+# By default, provide a simple help CMD and a direct ENTRYPOINT
+ENTRYPOINT ["geoips"]
+CMD ["geoips", "--help"]
```
jsolbrig, line 200: I think that we may want to consider having two completely different paths through this dockerfile. One that is intended for dev/test and one that is intended for production. We may also need some testing for production to ensure that things work when installed without `-e`. The `production` stage should be installed without `-e`.

This can be done in a follow-on issue but please create the issue and link it here.

Dockerfile
biosafetylvl5: Sure, that'd be fine with me - this results in a smaller image by running them all together but not by much

biosafetylvl5: IMO this is more of a problem with our install then with the dockerfile

jsolbrig (approved):
This still needs quite a bit of work to slim down the image and to incorporate both editable and non-editable installs. That said, this is a big step in the right direction and the rest should be handled in follow-on issues.

biosafetylvl5:
Re: your comments about testing, that is expected. You will need to run `base_test` instead of just `base` to have access to the test tooling.

biosafetylvl5:
Re: the question about the prod issue, the problem was on main. It should be working now.

Issue #1039, opened 2025-06-11
Add CLI flag to change warning level and default to printing warnings


jsolbrig:
# Requested Update

### Description
Update the CLI to include an option to change the warning "level". The option should be called `--warnings` and allow "hide", "print", and "fatal". These options would be used to set the warning level by calling `warnings.simplefilter(warning_level)`. When "hide", warning_level would be "ignore". When "print", warning_level would be "default". When "fatal", warning_level would be "error".

### Background and Motivation
Python, by default, does not show warnings when running a script or CLI. Warnings that are displayed when running interactively get suppressed when running a script or the CLI. It seems like better behavior if GeoIPS always prints warnings by default.

### Alternative Solutions
Another option would be to just hard-code `warnings.simplefilter("default")` but that wouldn't allow either suppression or error-on-warning.
  
### Checklist for Completion
* [ ] Add the new flag to GeoipsCommand._get_cli_logger().
* [ ] Rename _get_cli_logger() to something more appropriate and update its uses.
* [ ] Add the new flag to GeoipsCLI.__init__ in a similar way to how --log-level was added.

PR #1040, opened 2025-06-11
1039 add cli flag to change warning level and default to printing warnings

jsolbrig:
## ✨ Summary

Add a --warnings option to the CLI to allow controlling behavior when warnings are encountered. Can be set to hide, print, or raise an error on warnings.

---

---

## 🔗 Related Issues

- **Fixes:** #1039   
  

---

## 🧪 Testing Instructions

Try calling a test script with the different `--warnings` options and ensure that they behave as expected.


srikanth-kumar:
Optional:  `show` would be more user-intuitive than `print` as it better aligns with the opposite action `hide`. Suggesting this change/modification since  this command is likely to be a frequently used one.

Current: 

```
options:
  --warnings {hide,print,error}
```

Recommendation:

options:
  --warnings {hide,**show**,error}

srikanth-kumar:
> @srikanth-kumar I didn't even think about the `geoips test script` command. That adds a new wrinkle that I think I want to handle in a different PR. Most CLI commands will work as expected, including `geoips run`. `geoips test script` uses subprocess to call a test script, though, and has no means of passing arguments to that script.
> 
> One option that I'm thinking of to fix this is to have GeoIPS temporarily set "GEOIPS_WARN_LEVEL" or something like that.

That makes sense @jsolbrig. I think it would be better to handle this in a different PR; thank you for adding the rationale.

srikanth-kumar:
For a new `GeoIPS` user, the flag value `error` may not be intuitive. Do you want to add a brief description of what each option does. 

```
options:
  --warnings {hide,print,error}
   hide: suppress the warnings (more specifically all the warnings / specific type of warnings)
    print or show: 
    error

```

jsolbrig:
I decided to update this to default to `--warnings=hide` rather than `--warnings=print` because it is going to take a while to hammer out all of these warnings and they're pretty intrusive. I also added a GEOIPS_WARNING_LEVEL environment variable that can be set to "hide", "print", or "error".

srikanth-kumar (dismissed):

geoips/commandline/geoips_command.py
```diff
@@ -328,11 +345,22 @@ def _get_cli_logger(self):
             default="interactive",
             choices=["interactive", "debug", "info", "warning", "error"],
         )
+        independent_parser.add_argument(
+            "--warnings",
+            type=str,
+            default=os.getenv("GEOIPS_WARNING_LEVEL", "ignore"),
```
mindyls, line 351: Should this go in base_paths.py / future config ?  So we can better track what the various available settings / options are?

mindyls (approved):
I left one comment.

geoips/commandline/geoips_command.py
jsolbrig: Ah, probably. This was a last minute addition. I'll add this and will need reapproval.

mindyls (approved):

PR #1261, opened 2026-03-06
update dockerfile, helper utils for install, and test data downloading to enable self-hosted runners that run integration tests

biosafetylvl5:
## ✨ Summary

- [x] This replaces the install scripts with ansible (and patches the install scripts so they can still be called - now they just use ansible in the background).

It reworks the dockerfile to use the ansible scripts.

It adds a CI action that uses a selfhosted runner to run the integration tests.

It builds quick:

Screenshot 2026-06-25 at 3 50 21 PM

and runs the plugin integration tests as well:

Screenshot 2026-06-29 at 10 24 22 AM

Known issues:

1. Test data will stay stale up to a month - the config file PR fixes this
2. Non-editable tests fail (that's because of the code, not this PR) - we can make these non-blocking if we want

Unknown issues:
1. May not work well on plugin repos
2. May temporarily break doctest on plugin repos - no way to know until we merge to main though...

---

---
💖🌟🌎🌟💖


biosafetylvl5:
Commenting here!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,24 @@
+name: CI
```
mindyls: We are going to rename this integtest or something right (which includes doc tests and unit tests) ? This is only replacing doctest not all the ci?

Dockerfile
```diff
@@ -1,241 +1,300 @@
...
-       setup \
-       requirements.txt \
-    && apt-get remove -y \
-       git \
-       make \
-       g++ \
-       wget \
-       gfortran \
+# Only site-packages and console-script stubs — no source tree at all.
+COPY --from=geoips-base /usr/local/lib/python3.11/site-packages \
```
mindyls, line 305: Is this including only the base install? No plugin packages?

Makefile
```diff
@@ -0,0 +1,203 @@
...
+		.
+
+build-full:
+	DOCKER_BUILDKIT=1 docker build \
+		--target geoips-full \
+		--tag $(IMAGE_NAME):full-$(TAG) \
+		--build-arg EXTRA_PLUGINS="$(EXTRA_PLUGINS)" \
+		.
+
+build-site:
```
mindyls, line 41: All local builds are now docker based?

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+
+.. code-block:: text
+
+   deps       – pip install -r requirements.txt (cached layer)
+   geoips-base – ansible-playbook ... --tags base
+   geoips-full – ansible-playbook ... --tags base,full
+   geoips-site – ansible-playbook ... --tags base,full,site
+   production  – copy site-packages only, no source, no ansible, no git
+
+**Why ``pip_editable=false`` in Docker?**  The editable install mode makes the source
```
mindyls: There is an editable mode option though right for the docker dev?

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+ | # # # This source code is subject to the license referenced at
+ | # # # https://github.com/NRLMMD-GEOIPS.
+
+.. _ci-infrastructure:
+
+CI and Installation Infrastructure
+===================================
+
+GeoIPS uses `Ansible <https://docs.ansible.com/>`_ playbooks to manage installation of the
+software, its plugins, and test datasets.  The same playbooks run on bare-metal developer
```
jsolbrig: > GeoIPS uses Ansible ... test datasets.

GeoIPS uses Ansible ... test datasets for CI purposes.

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+
+
+Prerequisites
+-------------
+
+Install ``ansible-core`` into the same Python environment used for GeoIPS:
+
+.. code-block:: bash
+
+   pip install ansible-core
```
jsolbrig, line 60: This is a small, relatively quick install. Could this be done by the runners if not already available?

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+   :header-rows: 1
+   :widths: 10 90
+
+   * - Tag
+     - What it installs
+   * - ``base``
+     - Core GeoIPS package and plugin registries.  Sufficient for unit tests and
+       basic import checks.
+   * - ``full``
+     - Everything in ``base``, plus cartopy natural-earth shapefiles, settings repos
```
jsolbrig, line 100: Doesn't installing the shapefiles take a long time? Maybe these tests should be in a separate category? Or maybe we should update these tests to only use the low-res shapefiles?

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+Installing specific extra plugins
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Pass a comma-separated list of repository names:
+
+.. code-block:: bash
+
+   cd tests/ansible
+   ansible-playbook playbooks/install.yml --tags base \
+     -e extra_plugins=my_plugin,other_plugin
```
jsolbrig: Should this be "plugin packages" rather than "plugins"? If so, should `-e extra_plugins` be `-e extra_packages` or `-e extra_plugin_packages`?

Maybe it's enough to update the title of this section in the docs?

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+   :widths: 30 15 55
+
+   * - Variable
+     - Default
+     - Description
+   * - ``pip_editable``
+     - ``true``
+     - ``true`` uses ``pip install -e`` (editable — the source tree *is* the package).
+       ``false`` builds from source into ``site-packages``, which produces a smaller Docker
+       image.
```
jsolbrig, line 181: Is it possible for the CI to test in both editable and non-editable mode? We sometimes have problems where editable and non-editable make a difference. I think this is likely to continue until we fully transition to installing config and cache files in users' home directories.

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: sure, I'm fine with this change to the arguments (plugins to plugin packages)

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: Yes - this is replicating the existing CI with ansible. I agree that in later work we should not require the high-res shapefiles for `base` or even `full` !

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+   * - ``geoips_modified_branch``
+     - ``""``
+     - After cloning each repo, attempt to check out this branch.  Falls back silently to
+       the default branch if it does not exist.
+   * - ``geoips_packages_dir``
+     - ``/packages``
+     - Root directory where repos are cloned.  Reads ``GEOIPS_PACKAGES_DIR`` env var.
+   * - ``geoips_testdata_dir``
+     - ``/geoips_testdata``
+     - Root directory for test datasets.  Reads ``GEOIPS_TESTDATA_DIR`` env var.
```
jsolbrig, line 204: Maybe these defaults should come from the inventory? That way someone can use the playbooks locally and just create an inventory for their specific setup. I think this could be a follow-on issue.

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: Running the ansible workflows through pytest requires ansible-core. The runners install this. But if you're running locally you also need it as a pre-req (though it is also installed by the `test` extra.... so we could remove this if it's confusing).

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: Yeah, that's worth clarifying. I'll change that

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+
+
+Downloading test data
+----------------------
+
+Test data is managed by a **separate** playbook (``test_data.yml``) and is **never** baked
+into Docker images.  It uses the same tier tags as the install playbook.
```
jsolbrig, line 211: When this says that the test data are not baked into the docker images and talks about the installs being idempotent, does that mean that they are installed into a directory mounted from the host?

If that is right, do we have a mechanism for checking for updates for the test datasets? If not, we should create an issue to add a check to `geoips config install`.

jsolbrig: Okay, the answer to my question is, yes, they are installed to the local system. I don't think there is a mechanism yet for ensuring that these datasets are up to date. We should figure something out. I think we should keep this simple for now and assume that people are only testing the most recent version of GeoIPS.

In the future, I could see a need to tie versions of test datasets to versions of GeoIPS, but I don't think we should bother until it becomes an issue.

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+
+- **base**: ``test_data_amsr2``
+- **full**: ``test_data_amsub``, ``test_data_arctic_weather_satellite``, ``test_data_atms``,
+  ``test_data_cygnss``, ``test_data_fci``, ``test_data_gfs``, ``test_data_gpm``,
+  ``test_data_modis``, ``test_data_multi_scan_times``, ``test_data_saphir``,
+  ``test_data_sar``, ``test_data_scat``, ``test_data_seviri``, ``test_data_smap``,
+  ``test_data_smos``, ``test_data_tpw``, ``test_data_viirs``
+- **site**: ``template_test_data``, ``test_data_clavrx``, ``test_data_fusion``,
+  ``test_data_geocolor``
+- **private** (``geoips_use_private_plugins=true``): ``test_data_mint``
```
jsolbrig, line 266: Should this be in public documentation?

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+   CI scripts.  Only runs at ``full`` and ``site`` tiers.
+
+``source_repos``
+   Clones and ``pip install``\s plugin packages.  Handles four groups in order:
+
+   1. **Standard plugins** (alphabetical): ``data_fusion``, ``geoips_clavrx``,
+      ``geoips_plugin_example``, ``recenter_tc``, ``template_basic_plugin``
+   2. **Fortran chain** (order critical): ``fortran_utils`` → ``rayleigh`` → ``ancildat``
+      → ``synth_green`` → ``geocolor``
+   3. **Private plugins** (when enabled): ``ryglickicane``, ``tc_mint``
```
jsolbrig, line 308: Should this be in public documentation?

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+**Why ``pip_editable=false`` in Docker?**  The editable install mode makes the source
+directory itself the package.  In a container built for deployment this is undesirable —
+the source tree may not be present at runtime.  Building with ``pip_editable=false``
+installs the package into ``site-packages`` like a normal wheel, and the ``production``
+stage then copies only that directory, producing a minimal image.
+
+**Why ``--no-binary :all:``?**  Pre-built wheels are compiled for a generic architecture.
+Compiling from source inside the container produces binaries optimized for the target CPU
+and avoids wheel-cache bloat.  The ``deps`` stage also applies this flag, and because
+``requirements.txt`` changes infrequently, Docker caches that layer across most rebuilds.
```
jsolbrig, line 373: Is it appropriate to keep the wheel cache? I typically delete it after each `pip install` because it gets so big. I guess in a multi-stage built it doesn't make it into the final image, though, right? If so, does it actually get cached between stages for the next build? I think I'm a bit confused on how this would work here.

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+
+**Test data volume pattern:**
+
+.. code-block:: bash
+
+   # Download data to the host via the ansible playbook inside the container
+   make testdata-full TESTDATA=/data/geoips-testdata
```
jsolbrig: There likely needs to be a bit of reorganization or earlier explanation of the use of `make` in this process. This is the first mention of `make` so it comes as a surprise.

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+       ``--no-binary :all:`` to compile all dependencies from source inside the container.
+   * - ``geoips_use_private_plugins``
+     - ``false``
+     - Set to ``true`` to include proprietary plugin repos (``ryglickicane``, ``tc_mint``,
+       ``lunarref``, ``true_color``).
+   * - ``extra_plugins``
+     - ``""``
+     - Comma-separated list of additional plugin repository names to clone and install.
+       Corresponds to the ``EXTRA_PLUGINS`` Docker build argument.
+   * - ``geoips_modified_branch``
```
jsolbrig, line 194: This should provide a link to the "Branch fallback strategy" section. "For more information on how this variable is handled, please see the Branch fallback strategy section."

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+
+   # Run tests with that data mounted at the expected path
+   docker run --rm -v /data/geoips-testdata:/geoips_testdata geoips:full \
+     pytest -m "base and integration"
+
+Test data is always a runtime mount, never a build-time layer.
+
+
+Branch fallback strategy
+-------------------------
```
jsolbrig, line 390: If I'm understanding this correctly, the intent here is to allow installation of the correct branch from whatever repo we're trying to test without needing to specify which repo the specific branch should be installed from. So, just specifying this would result in most repos being on `main` and only repos that have the specified branch would be something other than `main`. Correct?

I definitely see how this would be useful for CI purposes, but I worry that this could be error prone for local use. It someone mistyped the branch name, it could result in using the wrong branch without even knowing it since this silently fails back to `main`. I'm not sure if this is actually an issue or not. 

I probably wouldn't worry about this for now, but it might be good to add the ability to override branches explicitly on a repo-by-repo basis and error when the specific branch isn't found. This is likely issue-worthy to at least investigate whether it's needed.

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+The ``source_repos`` role implements this per-repo in
+``roles/source_repos/tasks/clone_and_install.yml``.  The ``settings_repos`` role uses a
+batch approach: clone all repos in one loop with ``failed_when: false``, then re-clone only
+the ones that failed.
+
+
+Idempotency
+-----------
+
+The playbooks are designed to be re-run safely at any point:
```
jsolbrig, line 413: Maybe mention that this is more a feature of how Ansible works and less about how the playbooks were designed. Ansible is great about idempotency!

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+  overwrite uncommitted changes.
+- ``pip install`` with ``state: present`` is a no-op when the package is already installed
+  at the correct version.
+- The ``test_data`` role uses ``creates: "{{ geoips_testdata_dir }}/{{ item }}"`` so
+  Ansible skips the ``geoips config install`` call entirely when the dataset directory
+  already exists.
+- ``geoips config create-registries`` is idempotent by design.
+
+If a run fails partway through, fix the underlying issue and re-run the same command.
+Completed tasks will execute as fast no-ops.
```
jsolbrig: If a run fails partway through, fix the underlying issue and re-run the same command.
Completed tasks will execute as fast no-ops. ***Multiple runs will not hurt anything.***

docs/source/concepts/architecture/ci-infrastructure.rst
```diff
@@ -0,0 +1,456 @@
...
+
+
+Adding new repositories
+------------------------
+
```
jsolbrig, line 430: Some slight edits for clarity:

```
Repositories and which install group they belong to are managed by ``tests/ansible/group_vars/all.yml``. To add a new repo or move a repo to a different install group, edit ``tests/ansible/group_vars/all.yml`` as follows:

**Standard plugin repos** (no ordering constraint):
   Add the repo name to ``plugin_repos``

**Fortran plugin repos** (ordering constraint applies):
   Add the repo name to ``fortran_repos_ordered`` in the
   correct position in the dependency chain.  The comment in that file documents the
   required order.

**Private repos**:
   Add to ``private_plugin_repos`` or ``private_fortran_repos`` (for the fortran chain)

**Test datasets**:
   Test datasets are manage differently. Add the dataset name to the appropriate tier list in
   ``tests/ansible/roles/test_data/defaults/main.yml``.  The name must be recognized by
   ``geoips config install``.

File: docs/source/getting-started/installing/docker.rst [hunk 21]
```diff
@@ -0,0 +1,285 @@
...
+to install the software.  Each stage corresponds to an installation tier and
+can be targeted independently.
+
+.. contents:: On this page
+   :local:
+   :depth: 2
+
+
+Quick start
+-----------
```
jsolbrig, line 16: Please add a short description of what this quick start accomplishes. "Builds a docker image including GeoIPS Core and the packages installed in a `full` installation as described in the `ci-infrastructure` page. Then installs the required test data. Then ..."

Does running `make testdata-full` require having anything installed locally or should this be run from inside a container somehow?

docs/source/getting-started/installing/docker.rst
```diff
@@ -0,0 +1,285 @@
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 80
+
+   * - Target
+     - Description
+   * - ``base-os``
+     - System packages only (``git``, ``g++``, ``gfortran``, etc.).  Not
+       intended for direct use.
```
jsolbrig: Does this include ansible? If so, it seems important enough to include here. If not, where is ansible installed?

docs/source/getting-started/installing/docker.rst
```diff
@@ -0,0 +1,285 @@
...
+   * - Argument
+     - Default
+     - Description
+   * - ``EXTRA_PLUGINS``
+     - ``""``
+     - Comma-separated list of additional plugin repository names to clone
+       and install at any tier.
+   * - ``GEOIPS_MODIFIED_BRANCH``
+     - ``""``
+     - After cloning each repo, attempt to check out this branch.
```
jsolbrig, line 95: I think this deserves a bit more explanation. Things that are worth explaining:
1. If a particular repo does not have this branch, it will silently fail to check out this branch for that repo.
2. If multiple repos have the requested branch, the branch will be checked out for all of them.
3. This is intended to check out a particular set of changes to test against the `main` branch of all other relevant repos.

I'm not sure where this description should go. Maybe there is another section later?

docs/source/getting-started/installing/docker.rst
```diff
@@ -0,0 +1,285 @@
...
+     - Set to ``true`` at the ``geoips-site`` target to include private repos.
+   * - ``USER``
+     - ``geoips_user``
+     - Username for the non-root user inside the container.
+   * - ``USER_ID``
+     - ``1000``
+     - UID for the non-root user.
+   * - ``GROUP_ID``
+     - ``1000``
+     - GID for the non-root user.
```
jsolbrig, line 111: Why would someone want to change these?

docs/source/getting-started/installing/docker.rst
```diff
@@ -0,0 +1,285 @@
...
+
+The Dockerfile is structured to maximise Docker layer cache hits:
+
+1. **``base-os``** — Changes only when the base image or system packages
+   change (rare).
+
+2. **``deps``** — Copies *only* ``environments/requirements.txt`` and
+   installs all Python dependencies with ``--no-binary :all:``.  This is the
+   slowest layer but it rebuilds only when ``requirements.txt`` changes.
+   Day-to-day source edits skip this entirely.
```
jsolbrig, line 135: How and when does requirements.txt change? Is that a manual process at some point? I should know this, but I've generally just ignored requirements.txt. 😅

docs/source/getting-started/installing/docker.rst
```diff
@@ -0,0 +1,285 @@
...
+
+All Docker stages compile Python packages from source rather than using
+pre-built wheels.  This is controlled by two mechanisms:
+
+- The ``deps`` stage passes ``--no-binary :all:`` directly to ``pip install``
+  when installing ``requirements.txt``.
+
+- Each Ansible playbook invocation passes ``-e 'pip_extra_args=--no-binary
+  :all:'`` so that GeoIPS itself, its extras, and any plugin packages are
+  also compiled from source.
```
jsolbrig, line 157: It would be nice to clarify that the `-e` here is to pass an extra argument to ansible, not to perform an editable install via pip for people who are just scanning the docs.

docs/source/getting-started/installing/docker.rst
```diff
@@ -0,0 +1,285 @@
...
+       -i /packages/geoips/tests/ansible/inventory/local.yml \
+       --tags base,full \
+       -e geoips_testdata_dir=/geoips_testdata \
+       -v
+
+Or use the Makefile wrapper:
+
+.. code-block:: bash
+
+   make testdata-full TESTDATA=/data/geoips-testdata
```
jsolbrig: Does this specifically put the data in `/data/geoips-testdata` on the host? Does this have any prereqs like `make` or the GeoIPS repo? Do you need to be in a specific directory?

docs/source/getting-started/installing/docker.rst
```diff
@@ -0,0 +1,285 @@
...
+   make test-base  TESTDATA=/data/geoips-testdata
+   make test-full  TESTDATA=/data/geoips-testdata
+   make test-unit
+
+
+CI workflow
+-----------
+
+The GitHub Actions workflow (``.github/workflows/ci.yml``) automates the
+full cycle:
```
jsolbrig, line 257: Why? To build and push the image to dockerhub and for use in CI?

tests/ansible/group_vars/all.yml
```diff
@@ -0,0 +1,101 @@
+# ---------------------------------------------------------------------------
+# Paths – fall back to env vars (Docker sets them), then sensible defaults.
+# Override any of these with -e on the command line.
+# ---------------------------------------------------------------------------
+geoips_packages_dir: "{{ lookup('env', 'GEOIPS_PACKAGES_DIR') | default('/packages', true) }}"
+geoips_outdirs: "{{ lookup('env', 'GEOIPS_OUTDIRS') | default('/output', true) }}"
+geoips_testdata_dir: "{{ lookup('env', 'GEOIPS_TESTDATA_DIR') | default('/geoips_testdata', true) }}"
+geoips_dependencies_dir: "{{ lookup('env', 'GEOIPS_DEPENDENCIES_DIR') | default('/app/dependencies', true) }}"
+geoips_repo_url: "{{ lookup('env', 'GEOIPS_REPO_URL') | default('https://github.com/NRLMMD-GEOIPS', true) }}"
```
jsolbrig: Are these defaults appropriate? It seems easier to hard-code these into the docker/CI builds and error if they're not provided. This could be a tripping point for anyone who is using ansible locally. This is especially true because we don't expect users/developers to set some of these anymore (e.g. `GEOIPS_DEPENDENCIES_DIR` isn't required by our setup anymore).

tests/ansible/playbooks/test_data.yml
```diff
@@ -0,0 +1,88 @@
...
+  vars_files:
+    - ../group_vars/all.yml
+
+  tasks:
+    - name: Ensure GEOIPS_TESTDATA_DIR exists
+      ansible.builtin.file:
+        path: "{{ geoips_testdata_dir }}"
+        state: directory
+        mode: "0755"
+      tags: [base, full, site]
```
jsolbrig, line 37: Can we somehow check that the test data directory is on a volume with sufficient space? This can probably be an issue, especially since I'm not quite sure how to determine what "sufficient space" is depending on what is being installed.

tests/ansible/playbooks/test_data.yml
```diff
@@ -0,0 +1,88 @@
...
+    - name: Install base test datasets
+      ansible.builtin.command:
+        cmd: "geoips config install {{ item }}"
+        creates: "{{ geoips_testdata_dir }}/{{ item }}"
+      environment:
+        GEOIPS_TESTDATA_DIR: "{{ geoips_testdata_dir }}"
+      register: _td_base
+      changed_when: "'SUCCESS' in _td_base.stdout"
+      loop: "{{ base_test_datasets | default([]) }}"
+      tags: [base]
```
jsolbrig, line 49: Should the tags here be `[base, full, site]` so that the base test datasets always get installed? Similarly, should the tags on the "full" step below be `[full, site]`?

tests/ansible/roles/cartopy_shapefiles/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+# In Docker builds the shapefiles are pre-populated by the Dockerfile's deps
+# stage.  The block below is skipped when they already exist, keeping the
+# role usable for bare-metal installs.
+---
+- name: Check if shapefiles already exist (pre-populated by Dockerfile)
+  ansible.builtin.stat:
+    path: "{{ lookup('env', 'CARTOPY_DATA_DIR') | default(geoips_packages_dir, true) }}/shapefiles/natural_earth/cultural"
+  register: _shapefiles_exist
+  changed_when: false
+  tags: [full, site]
```
jsolbrig, line 21: These aren't need by `base`?

tests/ansible/roles/python_env/tasks/main.yml
jsolbrig, line 1: I'm a little confused by the tags in this file. We don't want to install linters when testing `base`?

tests/ansible/roles/registries/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+# roles/registries/tasks/main.yml
+---
+- name: Create GeoIPS plugin registries
+  ansible.builtin.command:
+    cmd: geoips config create-registries
+  changed_when: false
+  tags: [base, full, site]
```
jsolbrig, line 7: Should this include the `python_env` tag?

tests/ansible/roles/test_data/defaults/main.yml
jsolbrig, line 1: Does this need to be kept up to date the same way that `group_vars/all.yml` does?

tests/ansible/roles/test_data/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+    state: directory
+    mode: "0755"
+  tags: [base, full, site]
+
+# ---- base tier datasets ----
+- name: Install base datasets
+  ansible.builtin.include_tasks: install_datasets.yml
+  vars:
+    datasets: "{{ base_test_datasets }}"
+  tags: [base]
```
jsolbrig: Wouldn't this be required for `full` and `site` as well? Shouldn' this be `[base, full, site, python_env]`?

jsolbrig: Maybe the readme answers this question for me. I think the additive nature of these needs to be explained elsewhere. If the tags are additive, why do we need `[base, full, site, python_env]` in some places?

tests/ansible/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Tiers
+
+| Tag    | What it does                                                        |
+|--------|---------------------------------------------------------------------|
+| `base` | Installs GeoIPS core (`pip install -e .`), creates registries       |
+| `full` | Adds cartopy shapefiles, settings repos, doc/test extras            |
+| `site` | Adds all open-source plugin packages (fortran chain, geocolor, …)   |
+
+Tags are additive — always include all lower tiers:
```
jsolbrig, line 14: The fact that tags are additive should be mentioned in the rst docs, too. I was confused by this! Why do some tasks require `[base, full, site, python_env]` if the tags are additive? Wouldn't just `[base, python_env]` be sufficient in this case?

tests/ansible/roles/test_data/tasks/main.yml
biosafetylvl5: Well... technically we don't _need_ the `base` datasets for the `full` tests, no? Whereas some other things (like permissions above) _must_ be set when running tests from base/full/site/anything really, right?

biosafetylvl5: That's my thinking but it wouldn't really hurt anything to add full/site to this. It'd just no-op

Dockerfile
```diff
@@ -1,241 +1,300 @@
...
-    chmod -R 777 /__w /__e /__t /github /tmp # For github actions
+USER root
+RUN uv pip install --system --no-cache ${GEOIPS_PACKAGES_DIR}/geoips[doc,test] \
+    && cd ${GEOIPS_PACKAGES_DIR}/geoips/tests/ansible \
+    && ansible-playbook playbooks/install.yml \
+       --tags full \
+       --skip-tags python_env,cartopy_shapefiles \
+       -e pip_editable=false \
+       -v \
+    && chown -R ${USER_ID}:${GROUP_ID} ${GEOIPS_PACKAGES_DIR} /home/${USER}
```
jsolbrig, line 176: Is it appropriate to install as the root user here or would it be more appropriate to install as the requested user? Could this have security implications? Does this have implications for editability in devcontainers?

tests/ansible/roles/test_data/tasks/main.yml
jsolbrig: I suppose there could be a reason to run "full" without "base" but I've always assumed that if I'm running "site" that would include "full" and if I'm running "full" that would include "base". Maybe that's the wrong assumption?

jsolbrig (changes requested):
This looks great @biosafetylvl5! I left a bunch of comments, but they're all minor things and questions. Thank you for this! I'm really looking forward to automated integration testing!

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: Yes! There is a system!! If it's not in this PR, it's in my config file PR. We download the first few MB and hash that to see if things are up to date

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: Yes? It's already in the code at least... 👁️ 👁️

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: Again it's in the code .. ? @cpcamacho is this an issue?

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: This caches the built packages with the normal docker layer caching ? This is just saying its not using the pypi cached wheels and is instead building from source

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: Sure, I can move it up. Though honestly I don't really expect most (anyone) to use `make`, hence it's location

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: I can add that

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: That sounds reasonable to me - can you open an issue for that? I'm not 100% sure what you mean / what the solution/issue is. I don't really expect people to run this manually, and this is written mostly for the CI to use it. ATM I think I'm the only person using the docker image regularly even among the core developers.

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: It's really easy to write ansible playbooks that are not idempotent! Hence the wording here 🙂

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: Sure, I can add that

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: Sure, I'll copy and paste this in

docs/source/getting-started/installing/docker.rst
biosafetylvl5: Sure, I can add that. And no - the goal is that it does not require anything other than the prerequisites defined above

docs/source/getting-started/installing/docker.rst
biosafetylvl5: They are updated every release by mindy

docs/source/concepts/architecture/ci-infrastructure.rst
jsolbrig: Alright, thinking about it, this is fine. It's through public release so I'm not worried about it.

docs/source/getting-started/installing/docker.rst
biosafetylvl5: This is an example, so yes - in this example it puts it in /data/geoips-testdata on the host. I must be a little confused here? Yes? `make` requires having `make` installed? Can you clarify what the question is? Again - the makefile is only a convenience function for mimicking the CI when debugging

docs/source/getting-started/installing/docker.rst
biosafetylvl5: Why do we use a ci workflow? Or? Can you clarify the question?

tests/ansible/group_vars/all.yml
biosafetylvl5: I agree, but this is out of the scope of this PR as this PR replaces the shell scripts with identical functionality to the previously extant shell scripts. Can you make an issue to follow up on this and what you think should/shouldn't be required to be set?

tests/ansible/playbooks/test_data.yml
biosafetylvl5: No, we do not provide any way in our code to estimate the size of the external datasets. I agree this is something we should do though! Can you create an issue from this comment?

docs/source/concepts/architecture/ci-infrastructure.rst
jsolbrig: I can see myself running this manually instead of doing setup by hand. I used full_install.sh when it was working correctly for me, but it no longer does. I'm looking forward to being able to use ansible to set up environments for myself. Though, I may switch to devcontainers, which would mean I wouldn't use the ansible directly anymore.

docs/source/concepts/architecture/ci-infrastructure.rst
jsolbrig: Heh, okay, sounds good!

tests/ansible/roles/cartopy_shapefiles/tasks/main.yml
biosafetylvl5: No

tests/ansible/roles/python_env/tasks/main.yml
biosafetylvl5: We _could_ but there's no need to. Those should be in the python_env install set

tests/ansible/roles/test_data/tasks/main.yml
biosafetylvl5: I have _no_ idea why one would do that but theoretically? yes? you could?

docs/source/getting-started/installing/docker.rst
mindyls: There is also a requirements. Txt specifically for the complete site install if that helps

docs/source/getting-started/installing/docker.rst
mindyls: I see it requires make, and it must require geoips repo because that's where the makefile wrapper lives?

Maybe putting an obviously fake path would help Identify it is an arbitrary path on disk for the example? 

/arbitrary/path/to/dat/on/host/geoips-test-data

docs/source/getting-started/installing/docker.rst
```diff
@@ -0,0 +1,285 @@
...
+The downloaded data persists on the host.  Subsequent runs are fast because
+``geoips config install`` is idempotent.
+
+Run tests with mounted data
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: bash
+
+   docker run --rm \
+     -v /data/geoips-testdata:/geoips_testdata \
```
mindyls: I guess if you made the fake path really long you'd have to make it really long in all the examples..

/path/on/host/geoips-test-data

docs/source/getting-started/installing/docker.rst
```diff
@@ -0,0 +1,285 @@
...
+   docker run --rm geoips:prod --help
+
+Build-time tools (``git``, ``make``, ``g++``, ``gfortran``, ``wget``) are
+removed via ``apt-get remove`` to further reduce image size.
+
+
+Make targets
+------------
+
+The Makefile provides convenience wrappers:
```
mindyls: Maybe note where this makefile lives?

docs/source/getting-started/installing/docker.rst
mindyls: Are we going to have both a geoips only docker image and a complete site install docker image ?  So this will run

1. Geoips repo only integration tests
2. Plugin repo integration tests using geoips repo image
3. Complete integration tests across entire site install using the site install docker image ?

I can see this branch update thing being useful for integration testing the release branch across all repos during the release process ....

docs/source/releases/latest/skip-full-integration-low-ram.yaml
```diff
@@ -0,0 +1,22 @@
+continuous integration:
+- title: 'Skip full integration tests on low-RAM runners'
+  description: |
+    Adds a ``MemTotal`` check to the integration test step in the CI workflow.
+    If the host has less than 17 GB RAM (e.g. GitHub-hosted runners), the full
+    integration test suite is gracefully skipped with a log message. Base
+    integration tests continue to run everywhere.
```
mindyls: Are we sure base integration tests always use less than 16GB ram? Is this a standard that we need to enforce?

mindyls: Also don't we need to know how much memory we have,  and how much we need,  not just les than 16GB?

tests/ansible/group_vars/all.yml
mindyls: Won't hard coding paths break when running the CI on a different self hosted runner?  I would rather set CI variables and use those


This is gwyn - I cant (???) respond to this comment, but I can..... edit? Anyways..

Yes, you can set CI variables or ENV vars, this just sets defaults like the original dockerfile/scripts did

tests/ansible/group_vars/all.yml
```diff
@@ -0,0 +1,101 @@
...
+  - true_color
+
+# ---------------------------------------------------------------------------
+# Test datasets — which datasets belong to each install tier.
+# Each entry is a name recognised by `geoips config install <name>`.
+# These live here (not in role defaults) so they are always available
+# regardless of how the test_data playbook is invoked.
+# ---------------------------------------------------------------------------
+base_test_datasets:
+  - test_data_amsr2
```
mindyls: How are the test data downloads going to work with geoips config install-github? Is there an easy pace to swap out GeoipsConfigInstall for GeoipsConfigInstallGithub?

tests/ansible/playbooks/test_data.yml
```diff
@@ -0,0 +1,88 @@
...
+
+    # ---- full tier datasets ----
+    - name: Install full test datasets
+      ansible.builtin.command:
+        cmd: "geoips config install {{ item }}"
```
mindyls, line 54: Is this the only spot that needs to change for install-github? Will it just work?

mindyls (commented):
28/38

My main questions so far are:
1. Will geoips config install-github work just as well / the same as config install? 
2. If you remove the environment variables and hard code,  how will that impact running the CI on other self hosts runners ?

tests/ansible/roles/test_data/tasks/install_datasets.yml
```diff
@@ -0,0 +1,22 @@
...
+# Call via import_tasks with vars: {datasets: <list>}
+#
+# The `creates:` parameter lets Ansible skip the CLI call entirely when the
+# dataset directory already exists.  If a dataset is only partially extracted,
+# delete its directory to force a re-download.
+---
+
+- name: "Install test dataset: {{ item }}"
+  ansible.builtin.command:
+    cmd: "geoips config install {{ item }}"
```
mindyls, line 14: Would this work the same with install-github?

tests/ansible/roles/test_data/tasks/main.yml
mindyls: I would just asume full requires base, etc

mindyls (dismissed):
Please see my comments.  As long as geoips config install-github will work,  and we do not end up hard coding any paths that lock us into a specific self hosted runner,  this looks fine

docs/source/getting-started/installing/docker.rst
biosafetylvl5: Can do!

tests/ansible/group_vars/all.yml
biosafetylvl5: do you mean install datasets from github? IIRC the cli already supports that, they just have to be registered in github

docs/source/releases/latest/skip-full-integration-low-ram.yaml
biosafetylvl5: I based that off of the github self runners (which were able to run the `base` tests with 16GB of ram)

biosafetylvl5: so 16GB seemed sufficient for the base set

biosafetylvl5: but you could probably lower that without issue?

docs/source/getting-started/installing/docker.rst
biosafetylvl5: yeahh... but I can do that so it's clear

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: Yes there is!

Makefile
biosafetylvl5: it doesn't _have_ to be all, you can run the ansible locally

.github/workflows/ci.yml
biosafetylvl5: Yes! We should - this is just an easy name from me

tests/ansible/roles/test_data/tasks/main.yml
biosafetylvl5: Sure, that makes things the simplest

tests/ansible/roles/test_data/defaults/main.yml
biosafetylvl5: Yes it does, which isn't amazing but _is_ how the old scripts worked

tests/ansible/group_vars/all.yml
biosafetylvl5: We can set CI variables as env vars which would override these

docs/source/getting-started/installing/docker.rst
biosafetylvl5: but yes! we do want to push up the image for people to use for testing

docs/source/getting-started/installing/docker.rst
biosafetylvl5: ^ I'll switch this to an obvious fake path

biosafetylvl5: But yes it requires `make` and the GeoIPS repo with the makefile ? (is this answering the question ?)

docs/source/getting-started/installing/docker.rst
biosafetylvl5: Sure, I can add that

docs/source/getting-started/installing/docker.rst
biosafetylvl5: that does ! I think we can use that

docs/source/getting-started/installing/docker.rst
biosafetylvl5: some people have different user/group IDs and want non-root written files to be writable

docs/source/getting-started/installing/docker.rst
biosafetylvl5: Yes it should! I think it would really help to do so with the plugins for releases

docs/source/getting-started/installing/docker.rst
biosafetylvl5: hah, yes - it does include ansible :)

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: Yay! 😄

docs/source/getting-started/installing/docker.rst
jsolbrig: Yeah, I think that answers the question. Probably just update the path and close this conversation.

docs/source/getting-started/installing/docker.rst
jsolbrig: I'm not sure what I was asking here, so closing this :)

tests/ansible/roles/test_data/tasks/install_datasets.yml
biosafetylvl5: yes! I think so

tests/ansible/playbooks/test_data.yml
biosafetylvl5: Yes I think it should just work - that or you need to change/update this section

docs/source/releases/latest/skip-full-integration-low-ram.yaml
biosafetylvl5: so not a standard, more of a "oh neat! that works!" if things start failing we can either turn things off / update the tests and make that call at the time IMO

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: fixed!

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: fixed!

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: Yes, but that's going to have to be future work.... almost none of the test sets work in non-editable mode right now.

tests/ansible/playbooks/test_data.yml
biosafetylvl5: ^^ I really think we need to overhaul our test data metadata (or... rather... get some metadata...)

tests/ansible/roles/registries/tasks/main.yml
biosafetylvl5: Why?

biosafetylvl5: I don't think so?

Dockerfile
biosafetylvl5: We probably could install as the user, but since we chown it shouldn't have any implications for dev containers

.github/workflows/ci.yml
biosafetylvl5: changed!

Makefile
biosafetylvl5: it's basically just a wrapper for the bash

Dockerfile
biosafetylvl5: yes! for the production image I only copy over the base install

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: moving!

tests/ansible/README.md
biosafetylvl5: Yes, but only base, full, and site are additive - not python_env. That's how all of the test currently work on `main`

tests/ansible/playbooks/test_data.yml
biosafetylvl5: I don't think so? it goes base -> full -> site in scope

docs/source/getting-started/installing/docker.rst
biosafetylvl5: updated

docs/source/getting-started/installing/docker.rst
biosafetylvl5: updated

biosafetylvl5:
Okay - as far as I can tell all the comments have been addressed!

.github/workflows/integ_test.yaml
```diff
@@ -0,0 +1,32 @@
...
+  packages: write
+
+jobs:
+  integ-test:
+    if: github.event_name == 'workflow_dispatch' || github.event.pull_request.draft == false
+    strategy:
+      matrix:
+        pip_editable: [true, false]
+    name: integ-test (editable=${{ matrix.pip_editable }})
+    uses: NRLMMD-GEOIPS/geoips_ci/.github/workflows/reusable-ci.yaml@bsl5_ci_fix
```
mindyls: We need to update this before merge?

.github/workflows/status-check.yaml
```diff
@@ -11,6 +12,7 @@ permissions:
 
 jobs:
   status-check:
+    if: github.event_name == 'push' || github.event.pull_request.draft == false
```
mindyls, line 15: Will this not run on draft pull requests?

docs/source/concepts/architecture/ci-infrastructure.rst
mindyls: Create issue?

docs/source/concepts/architecture/ci-infrastructure.rst
mindyls: I think we will always need to test but editable and non editable.  You added both right?  The non editable just fail currently?  Create issue to fix the non editable install s?

docs/source/concepts/architecture/ci-infrastructure.rst
mindyls: So any self hosted runner will need local test data. Any other runner requirements?

tests/ansible/playbooks/test_data.yml
mindyls: Further review comment?

.github/workflows/integ_test.yaml
jsolbrig: This needs to be set to `main`. I'll take care of that in a bit.

docs/source/concepts/architecture/ci-infrastructure.rst
jsolbrig: Eh, it's fine :)

docs/source/concepts/architecture/ci-infrastructure.rst
jsolbrig: They normally work for me in non-editable mode, at least for core GeoIPS.

tests/ansible/playbooks/test_data.yml
jsolbrig: I think we should write a script that uploads/updates test data on io2 and generates and uploads an accompanying metadata file. The metadata file could serve as our token and include date generated, size, and maybe an associated GeoIPS version.

@mindyls you can resolve this comment when you get to it.

tests/ansible/playbooks/test_data.yml
jsolbrig: I'm going to assume you're right on this unless proven otherwise 😄

tests/ansible/roles/test_data/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+  vars:
+    datasets: "{{ full_test_datasets }}"
+  tags: [base, full, site]
+
+# ---- site tier datasets ----
+- name: Install site datasets
+  ansible.builtin.include_tasks: install_datasets.yml
+  vars:
+    datasets: "{{ site_test_datasets }}"
+  tags: [base, full, site]
```
jsolbrig, line 42: I'm wondering, for sections like this, would it be easier to put the conditional here rather than on every task in `install_datasets.yaml`? This works fine but it seems like `install_datasets.yaml` could be shortened quite a bit by removing all of the conditionals and setting them here.

Dockerfile
jsolbrig: Eventually, maybe we should supply a few different combinations. Something like base, full, and site but maybe a few more. This would probably come after doing a plugin reorganization.

.github/workflows/integ_test.yaml
jsolbrig: Fixed.

jsolbrig (dismissed):
This looks great! Thank you so much for working on this @biosafetylvl5! It is going to be amazing to not need to always manually run the integration tests!

tests/integration_tests/test_integration.py
```diff
@@ -33,8 +41,22 @@
...
-    "$geoips_repopath/tests/utils/check_code.sh all $repopath",
-    "$geoips_repopath/docs/build_docs.sh $repopath $pkgname html_only",
+    (
+        "pytest -q --tb=short"
+        " -m 'lint and not doc8'"
+        " $geoips_repopath/tests/lint/"
+        " --lint-path $repopath"
+    ),
+    (
+        "python3 $GEOIPS_PACKAGES_DIR/geoips/docs/build_docs.py"
```
mindyls, line 57: Is this exactly the same as build_docs.sh was?

tests/integration_tests/test_integration.py
```diff
@@ -33,8 +41,22 @@
 
 # Linting integration tests, ensure code and documentation are correctly formatted.
 lint_integ_test_calls = [
-    "$geoips_repopath/tests/utils/check_code.sh all $repopath",
-    "$geoips_repopath/docs/build_docs.sh $repopath $pkgname html_only",
+    (
+        "pytest -q --tb=short"
+        " -m 'lint and not doc8'"
+        " $geoips_repopath/tests/lint/"
```
mindyls, line 53: Does this do exactly the same thing as check cid did?

tests/integration_tests/test_integration.py
```diff
@@ -158,176 +180,175 @@ def set_log_filename(expanded_call):
     return log_fname
 
 
-def check_environment(check_environment_script):
```
mindyls, line 161: Can we put this back? I don't think it hurt anything?

tests/integration_tests/test_integration.py
```diff
@@ -158,176 +180,175 @@ def set_log_filename(expanded_call):
...
+    setting environment variables required for the integration tests.  Assumes
+    that ``GEOIPS_PACKAGES_DIR`` is already set in the environment.
 
     Notes
     -----
     The following environment variables are set:
+
     - geoips_repopath
     - geoips_pkgname
+    - repopath (alias for geoips_repopath — used by lint scripts)
```
mindyls: These are different for plugin repos right? geoips_repopath used for all plugin repos as the path to main geoips repo, repopath is current plugin repo?

tests/integration_tests/test_integration.py
```diff
@@ -158,176 +180,175 @@ def set_log_filename(expanded_call):
...
+    - pkgname (alias for geoips_pkgname — used by lint scripts)
     """
-    # Base path
-    os.environ["geoips_repopath"] = os.path.realpath(
-        os.path.join(os.path.dirname(__file__), "..", "..")
-    )
+    os.environ["geoips_repopath"] = _REPO_ROOT
     os.environ["geoips_pkgname"] = "geoips"
 
-    # Setup current repo's environment
```
mindyls, line 307: This is still the current repo,  it just happens to be geoips repo

tests/integration_tests/test_integration.py
```diff
@@ -351,38 +372,30 @@ def is_likely_oserror_missing_file(log_contents: str) -> bool:
...
 
     Raises
     ------
     subprocess.CalledProcessError
         If the shell command returns a non-zero exit status.
     """
-    # The print statements all appear to print AFTER the script is complete
-    # And appears to not include a newline after printing the script.
     print("")
-    # Unset all supported specific output path env vars to ensure consistent test
```
mindyls, line 377: This is all still relevant

tests/integration_tests/test_integration.py
```diff
@@ -351,38 +372,30 @@ def is_likely_oserror_missing_file(log_contents: str) -> bool:
...
+        If False, xfail the test instead.
+    unset_output_path_env_vars : bool, optional
+        If True, unset output path env vars for test consistency.
 
     Raises
     ------
     subprocess.CalledProcessError
         If the shell command returns a non-zero exit status.
     """
-    # The print statements all appear to print AFTER the script is complete
```
mindyls, line 374: This is still relevant

tests/integration_tests/test_integration.py
```diff
@@ -431,26 +438,40 @@ def run_script_with_bash(
...
-                print("\nLOG FILE CONTENTS:\n")
-                print(log_contents)
-                print("-" * 80)
+        stdout_str = "".join(stdout_list)
+        stderr_str = "".join(stderr_list)
+        combined = (stdout_str + stderr_str).strip()
+        summary_lines = combined.splitlines()[-25:]
+        summary = "\n".join(summary_lines)
+
+        if not gpaths["GEOIPS_TEST_SUPPRESS_PYTEST_FAILED_LOG_CONTENTS"]:
```
mindyls, line 487: Does this do the same thing still?

tests/lint/conftest.py
```diff
@@ -0,0 +1,62 @@
...
+Provides ``--lint-path`` and ``--flake8-docstring-only`` CLI options,
+registers lint-specific markers, and exposes the ``lint_path`` and
+``geoips_root`` fixtures.
+"""
+
+import os
+import pytest
+
+
+def pytest_addoption(parser):
```
mindyls, line 12: This only needs to be defined in one repo and called from all plugin repositories?

mindyls (commented):
45/45

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: I think we need to make the geoips tests (at least) installable + refactor all of the plugin tests to make this work

docs/source/concepts/architecture/ci-infrastructure.rst
biosafetylvl5: If you have <16gb of RAM, it'll only run the base tests - otherwise it'll run the full suite. So all the usual CPU and RAM requirements apply

Dockerfile
biosafetylvl5: Agreed - this is the set mindy and I landed on initially, but there are more tags we want to add.... eg. GEOs, LEOs, Model_data, etc..

biosafetylvl5: and by _we_ that might just be _me_

tests/ansible/roles/test_data/tasks/main.yml
biosafetylvl5: It probably could, but since folks (except you and I) are very new to ansible I tried to make `main` as simple as possible... even if that meant more complexity a bit deeper in. 

A judgement call that I'm happy to go either way with

.github/workflows/integ_test.yaml
biosafetylvl5: Yes we do!

.github/workflows/status-check.yaml
biosafetylvl5: Yes, it will not unless you manually dispatch it

tests/integration_tests/test_integration.py
biosafetylvl5: Yes! I just copied it over

tests/integration_tests/test_integration.py
biosafetylvl5: yep!

tests/lint/conftest.py
biosafetylvl5: Yes - it just needs imported by all the plugin repos

mindyls (dismissed):

.github/workflows/integ_test.yaml
biosafetylvl5: updated!

tests/integration_tests/test_integration.py
biosafetylvl5: Yes it just catches a few more of the edge cases

tests/integration_tests/test_integration.py
biosafetylvl5: I want to update this with python that calls the ansible... which is why it was removed. I haven't implemented that yet, I'll create an issue to do that after the workshop

tests/integration_tests/test_integration.py
biosafetylvl5: gosh, yes you're right. I need to fix the linting scripts - I was only thinking of the GeoIPS repo for this one

tests/integration_tests/test_integration.py
biosafetylvl5: ... let me see if this code is imported by the other repos test modules

tests/integration_tests/test_integration.py
biosafetylvl5: ag, rouge linter. Let me put this back

tests/integration_tests/test_integration.py
biosafetylvl5: ag, rouge linter. Let me put this back

tests/integration_tests/test_integration.py
biosafetylvl5: oh wait! I did!

biosafetylvl5: It didn't hurt, but I replaced with the same but better functionality

biosafetylvl5: which should still be running! 🙂

tests/integration_tests/test_integration.py
biosafetylvl5: but your right that for plugin repos the repopath is not geoips_repopath and the integration test dir may change when the integration test functions are imported into other packages

tests/ansible/playbooks/test_data.yml
biosafetylvl5: Yes, that needs added here just for you to check to make sure it works

tests/integration_tests/test_integration.py
biosafetylvl5: fixed

tests/integration_tests/test_integration.py
biosafetylvl5: fixed

tests/integration_tests/test_integration.py
```diff
@@ -85,7 +91,16 @@
     "$geoips_repopath/tests/scripts/ami.static.Infrared.imagery_clean.sh",
     "$geoips_repopath/tests/scripts/ami.static.Visible.imagery_clean.sh",
     "$geoips_repopath/tests/scripts/ami.static.mst.absdiff-IR-BD.imagery_clean.sh",
-    "$geoips_repopath/tests/scripts/ami.tc.WV.geotiff.sh",
+    # Temporarily make this xfail for github actions runs
+    # "$geoips_repopath/tests/scripts/ami.tc.WV.geotiff.sh",
+    pytest.param(
+        "$geoips_repopath/tests/scripts/ami.tc.WV.geotiff.sh",
+        marks=pytest.mark.xfail(
+            # condition=IS_ARM,
```
mindyls, line 99: We will have to uncomment this at some point?

mindyls (dismissed):

mindyls (dismissed):

tests/ansible/playbooks/test_data.yml
mindyls: Created an issue:
* https://github.com/NRLMMD-GEOIPS/geoips/issues/1408

docs/source/releases/latest/skip-full-integration-low-ram.yaml
mindyls: Ok, I am more concerned about plugin packages though, not the geoips repo.

mindyls: Created an issue: 
* https://github.com/NRLMMD-GEOIPS/geoips/issues/1409

tests/integration_tests/test_integration.py
jsolbrig: At some point, maybe. There are two things that need to happen.
1. We need to add some environment variables to our docker container at runtime that tells us something about what machine and architecture we're running on. If that's there, then I can start testing whether the chips are ARM or x86. At the moment, the info doesn't exist inside the docker container so I can't test what the chip architecture is. See https://github.com/NRLMMD-GEOIPS/geoips_ci/issues/80
2. We just need to get these tests passing on ARM. 

One nice note is that last year we were struggling to install on ARM and this year we're running our CI on ARM 🥳

jsolbrig (dismissed):

mindyls (commented):
47/47

tests/ansible/roles/python_env/tasks/main.yml
mindyls: I view the Base tests as just masking sure the package runs at all,  minimum requirements beyond just running python.  So I would think Base would *not* include linters. But, when we pip install with test, it installs all the linters, we just don't run them with base. 

Is this correct ?  Install liners for all tests (included in python_env),  just don't run them in base?

mindyls (commented):
47/47

mindyls (commented):
47/47

tests/ansible/roles/test_data/tasks/main.yml
mindyls: Is this outdated? I don't see any conditionals here or in install datasets?  And they both seem pretty straightforward

tests/ansible/roles/test_data/defaults/main.yml
mindyls: Isn't it OK to have to keep the list of required test data repos up to date for each testing tier? Was there something duplicated in the old installation scripts with test data repo lists?

Dockerfile
mindyls: So if I'm tracking properly,  effectively for CI purposes there are 3 explicit base, full, and site tags that are referenced in the main geoips repo within:
* Makefile
* Dockerfile
* ansible playbooks

And the geoips-base, geoips-full, and geoips-site tags must be explicitly referenced in each of those locations,  with the explicit list of plugin packages / test datasets included in the ansible playbooks ?

And you are talking about having additional combinations of packages "pre-loaded" under additional different tags (which would then be explicitly added to the Makefile, Dockerfile, and Ansible playbooks? )? 

I think more beneficial than that would be to make it easy for people to add their own tags / playbooks to install their own particular set of plugin packages that they are interested in - is that planned for the future? Actual functional installs are going to be very user specific,  so I think trying to predict useful combinations of plugin packages might just be more confusing and less helpful than just explicitly listing your own list of packages ?

tests/integration_tests/test_integration.py
mindyls: What is still running?  How will this change impact the plugin repositories integration test scripts?  If a plugin repo was calling check environment with a specific "check_environment_script" how would that functionality be replicated now ? Does this relate to my question about custom ansible playbooks / tags ?  In order for a custom environment to be "checked" in the current set up,  more work needs to be done to generalize the ansible playbooks / Makefile / Dockerfile?

tests/ansible/roles/test_data/tasks/main.yml
jsolbrig: It's not. By "conditionals" I mean adding tags to every task. These could be relocated to `install_datasets.yml` and have the same effect. That said, this is fine so I"m going to resolve this comment.

tests/ansible/roles/python_env/tasks/main.yml
biosafetylvl5: That's right, that's my thought process as well. So `python_env` is the correct place for the linters (which is where they are currently)

tests/ansible/roles/test_data/defaults/main.yml
biosafetylvl5: I think it's 100% OK, ideally I wish we put these in the `pyproject.toml` or in pytests so we could automatically know what plugins/tests need what BUT I am 11/10 good with the current solution

Dockerfile
biosafetylvl5: I think the `tags` we're talking about here are actually better placed as `extras` in the toml. Ideally we can keep the dockerfile, makefile and ansible similar to where they are now! And then others can write their own dockerfiles that inherit from ours or write their own playbooks if they have specific installs

biosafetylvl5: @mindyls does that work?

tests/integration_tests/test_integration.py
biosafetylvl5: Sorry, that wasn't very clear... The ansible checks should still be running what the check env script used to do - so not really playbooks/tags, but yes - in order for a custom env to be "checked" we'd need to do more work. Do we need that functionality? You can add/remove plugins already from the ansible books with env vars if that helps

jsolbrig (approved):

christopher-selman (approved):

PR #1262, opened 2026-03-07
OBP: Pydantic Models: Algorithms; Procflow Support: Algorithms

srikanth-kumar:
## ✨ Summary

This pull request adds a Pydantic-based `AlgorithmArgumentsModel` implementation for the Order-Based Procflow (OBP).
- Adds a new `AlgorithmStepValidationModel`  to validate step-level requirements for algorithm plugin.
  - Adds  a model validator to check if the algorithm inputs for specific plugins contains `variables`
- 73 unit tests providing 100% coverage for multiple pydantic models inherited by the primary arguments model. 

---

---

## 🧪 Testing Instructions

Run `pytest`  in `geoips/geoips/tests/unit_tests/pydantic_models/v1/`

image

---

💖🌟🌎🌟💖


tests/unit_tests/pydantic_models/v1/conftest.py
```diff
@@ -142,9 +142,23 @@ def valid_title_formatter_arguments():
...
+def valid_algorithm_arguments():
+    """Fixture providing valid data for AlgorithmArgumentsModel."""
+    return {
+        "output_data_range": [-90.0, 30.0],
+        "input_units": "Kelvin",
+        "output_units": "celsius",
+        "min_outbounds": "crop",
+        "max_outbounds": "crop",
+        "norm": None,
+        "inverse": None,
```
srikanth-kumar: @jsolbrig @mindyls Currently, SSP workflows ([reference](https://github.com/NRLMMD-GEOIPS/geoips/blob/main/geoips/plugins/yaml/product_defaults/visir/Infrared.yaml)) accepts `null` as a valid input for `norm` and `inverse`.

In the `AlgorithmArugmentsModel`, the type for both of these fields / members is `bool` (algorithm call signature, line numbers: 31 & 32 [reference](https://github.com/NRLMMD-GEOIPS/geoips/blob/main/geoips/plugins/modules/algorithms/model/model_channel.py)). Typically, a boolean variable should not accept `null` and fail as shown in the image below.
image

Are we okay with updating the references in the SSP workflows including products and product defaults to correct this (recommended) or should this be accommodated via a field validator for now?

jsolbrig: I'm of the opinion that we shouldn't be accepting `Null` anywhere. I think this is a holdover from how jsonschema worked where, if a value wasn't provided but was part of the schema, it would default to `Null`. I'm fully in support of changing this.

srikanth-kumar: I agree with, Jeremy, on not accepting `null` anywhere. For now, to keep the SSP->OBP conversion seamless, I have made these changes only to boolean types in this PR itself. The remaining cases will be addressed after the SSP->OBP implementation is complete.

srikanth-kumar:
This PR is ready for your review @jsolbrig @mindyls

geoips/plugins/modules/procflows/order_based.py
```diff
@@ -37,7 +37,7 @@ def call(workflow, fnames, command_line_args=None):
     LOG.interactive(f"Begin processing '{workflow['name']}' workflow.")
     wf_plugin = workflow
 
-    handled_interfaces = ["readers", "coverage_checkers"]
+    handled_interfaces = ["readers", "coverage_checkers", "algorithms"]
```
mindyls: So we are now up to officially supporting readers, coverage_checkers, and algorithms?

srikanth-kumar: Others as well such as `title_formatters`, `gridline_annotators`, `coverage_checkers`, `feature_annotators`, `output_checkers`, `products`, and `sectors` but these are not fully tested for all the products or scripts in the `test/scripts`. 

This line is updated in each OBP PR to evaluate if the corresponding plugin is getting called or not.

mindyls (commented):
76/81

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,143 @@
...
+    variables: Optional[List[str]] = Field(
+        [None],
+        description="List of input variables used in algorithm processing, selects the "
+        "first when provided.",
+    )
+    output_data_range: Optional[tuple[StrictFloat, StrictFloat] | tuple[None, None]] = (
+        Field(
+            [None, None],
+            description="list of min and max value for wind speeds (kts or m s-1). "
+            "Defaults to None, which results in using data.min and data.max.",
```
mindyls: Is this explicitly always wind speeds ?

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,143 @@
...
+
+    * Selection of data along time or other dimensions
+    * Optional scaling of output data variables
+    * Processing of values outside the specified data range
+    """
+
+    variables: Optional[List[str]] = Field(
+        [None],
+        description="List of input variables used in algorithm processing, selects the "
+        "first when provided.",
```
mindyls: This won't apply to all plugins

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,143 @@
...
+        description="Boolean flag indicating whether to normalize (True) or not (False)"
+        "* * If True, returned data will be in the range from 0 to 1:"
+        "  * If False, returned data will be in the range from min_val to max_val",
+    )
+    inverse: StrictBool = Field(
+        False,
+        description="* Boolean flag indicating whether to inverse (True) or not (False)"
+        " * If True, returned data will be inverted"
+        " * If False, returned data will not be inverted",
+    )
```
mindyls: I think all the arguments shuld be optional.

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,143 @@
...
+        "necessary conversions. Defaults to None, resulting in no unit conversions.",
+    )
+
+
+class PressureWindsAlgorithmArgumentsModel(FrozenModel):
+    """Arguments specific to Dervied Motion Wind Products."""
+
+    var_map: Optional[Dict[str, str]] = Field(
+        {}, description="Dictionary that maps input variables to names used in xobj"
+    )
```
mindyls: Should this be specific to PressureWinds ?

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,143 @@
...
+    """Arguments specific to Visible and Infrared algorithm plugins."""
+
+    gamma_list: Optional[List[StrictFloat]] = Field([])
+    min_night_zen: Optional[float] = Field(None)
+    max_night_zen: StrictFloat = Field(90)
+    max_day_zen: Optional[float] = Field(None)
+    mask_night: bool = Field(False)
+    mask_day: bool = Field(False)
+    scale_factor: Optional[float] = Field(None)
+    sun_zen_correction: StrictBool = Field(
```
mindyls: Are all of these specific to vis/ir?

mindyls (commented):
77/81

mindyls (commented):
81/81

mindyls (commented):
81/81

geoips/pydantic_models/v1/algorithms.py
jsolbrig: No, this description should be updated from single_channel I think.

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+
+    variables: Optional[List[str]] = Field(
+        [None],
+        description="List of input variables used in algorithm processing, selects the "
+        "first when provided.",
+    )
+    output_data_range: Optional[tuple[StrictFloat, StrictFloat] | tuple[None, None]] = (
+        Field(
+            [None, None],
```
jsolbrig: No default here.

geoips/pydantic_models/v1/algorithms.py
jsolbrig: Jeremy, look into how this is used. I think this is specifically used in the pre_call hook and probably IS required for most algorithms, but I'm not sure if it's always required. This will mean looking at single_source to figure out how `variables` is used.

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,143 @@
...
+            description="list of min and max value for wind speeds (kts or m s-1). "
+            "Defaults to None, which results in using data.min and data.max.",
+        )
+    )
+    min_outbounds: str = Field(
+        ...,
+        description="Method to use when applying bounds."
```
jsolbrig: Method to use when applying minimum value of `output_data_range`, if specified.

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,143 @@
...
+        "  Valid values are: "
+        " * retain: keep all pixels as is"
+        " * mask: mask all pixels that are out of range"
+        " * crop: set out of range values to the nearest bound (min_val or max_val)",
+    )
+    max_outbounds: str = Field(
+        ...,
+        description="Method to use when applying bounds."
```
jsolbrig: Method to use when applying maximum value of `output_data_range`, if specified.

geoips/pydantic_models/v1/algorithms.py
srikanth-kumar: Thanks for the correction, Jeremy. Updated and this comment is ready to be resolved.

geoips/pydantic_models/v1/algorithms.py
srikanth-kumar: Thanks for the correction, Jeremy. Updated and this comment is ready to be resolved.

geoips/pydantic_models/v1/algorithms.py
srikanth-kumar: Merged into a single model for now as suggested by @jsolbrig. We will be revisiting this post workshop.

geoips/pydantic_models/v1/algorithms.py
srikanth-kumar: Merged into a single model for now as suggested by @jsolbrig. We will be revisiting this post workshop.

geoips/pydantic_models/v1/algorithms.py
srikanth-kumar: Thanks, Mindy, for the correction. Updated as suggested.

geoips/pydantic_models/v1/algorithms.py
srikanth-kumar: Thanks for the correction, Mindy.

Corrected to: List of input variables used in algorithm processing"

geoips/pydantic_models/v1/algorithms.py
srikanth-kumar: Thank you, @mindyls  for the question and @jsolbrig for the recommendation.
The single_channel description makes more sense. 

Updated to: 

```
    description="list of min and max value for output data product."
        " This is applied LAST after all other corrections/adjustments."
        " If None, use data.min() and data.max()",
 ```

File: geoips/pydantic_models/v1/algorithms.py (hunk 8 above)
Comment (srikanth-kumar) [reply geoips/pydantic_models/v1/algorithms.py]: Updated to no default.

File: geoips/plugins/modules/filename_formatters/geoips_fname.py [hunk 11]
```diff
@@ -214,8 +215,12 @@ def assemble_geoips_fname(
...
         ]
     )
-    fname = "{0}.{1}".format(fname, output_type)
-    return pathjoin(path, fname)
+    fname = f"{fname}.{output_type}"
+    file_path = pathjoin(path, fname)
+    if path_exists(file_path):
+        return file_path
+    else:
+        raise FileNotFoundError(f"file does not exist: {file_path}")
```
mindyls: Why are we erroring if the file doesn't already exist?  We produce the output file AFTER we determine the filename we want to produce. Won't this result in us always failing and never producing anything?

tests/scripts/seviri.WV-Upper.no_self_register.unprojected_image.sh
```diff
@@ -17,3 +17,8 @@ geoips run single_source $GEOIPS_TESTDATA_DIR/test_data_seviri/data/20250624/120
 retval=$?
 
 exit $retval
+
+    # extension = output_type.lstrip(".")
+    # fname = f"{fname}.{extension}"
+    # full_fname = pathjoin(path, fname)
+    # return full_fname
```
mindyls: What are these comments used for?

mindyls: Oh! I just confirmed this WV-Upper test output DOES have the double ".."!  Make sure whenever we remove the double dot from the basic_fname filename formatter that we also update the WV-Upper test output file to remove the extra .

geoips/plugins/modules/filename_formatters/geoips_fname.py
srikanth-kumar: Lines 220 - 223 are invalid in this case since the file_path generated here is just string. 

Thanks for the correction, Mindy. These changes should not have been in the first place as they are out of the scope of this PR (likely from another patch branch) .

Removed.

geoips/plugins/modules/filename_formatters/basic_fname.py
```diff
@@ -45,5 +46,9 @@ def call(
...
         ]
     )
-    fname = "{0}.{1}".format(fname, output_type)
-    return pathjoin(path, fname)
+    fname = f"{fname}.{output_type}"
+    file_path = pathjoin(path, fname)
+    if path_exists(file_path):
+        return file_path
+    else:
+        raise FileNotFoundError(f"file does not exist: {file_path}")
```
mindyls: All of these need to be backed out.

geoips/plugins/modules/filename_formatters/geoips_netcdf_fname.py
```diff
@@ -120,4 +121,8 @@ def assemble_geoips_netcdf_fname(
         path_parts.extend([sector_name])
     fname = ".".join(path_parts + ["nc"])
 
-    return pathjoin(path, fname)
+    file_path = pathjoin(path, fname)
+    if path_exists(file_path):
+        return file_path
+    else:
+        raise FileNotFoundError(f"file does not exist: {file_path}")
```
mindyls: Revert please

geoips/plugins/modules/filename_formatters/tc_fname.py
```diff
@@ -407,5 +408,9 @@ def assemble_tc_fname(
...
         ]
     )
-    fname = "{0}.{1}".format(fname, output_type)
-    return pathjoin(path, fname)
+    fname = f"{fname}.{output_type}"
+    file_path = pathjoin(path, fname)
+    if path_exists(file_path):
+        return file_path
+    else:
+        raise FileNotFoundError(f"file does not exist: {file_path}")
```
mindyls: Revert please.

geoips/plugins/modules/filename_formatters/basic_fname.py
srikanth-kumar: Done, sorry about that, these are changes from the patch branch.

Thank you for catching this.

geoips/plugins/modules/filename_formatters/geoips_netcdf_fname.py
srikanth-kumar: Done

geoips/plugins/modules/filename_formatters/tc_fname.py
srikanth-kumar: Done.

mindyls (dismissed):

tests/scripts/seviri.WV-Upper.no_self_register.unprojected_image.sh
srikanth-kumar: Thanks a lot, Mindy! This is fixed.

mindyls (dismissed):

srikanth-kumar:
@jsolbrig @mindyls changes from the last review:

- After merging the changes from `main` there were 21 test failures because of the required vs optional config settings. This has been fixed making most of the fields optional (by defaulting to `None`).
- For fields `inverse` and `norm` the default values is set to `False`. 
- I've also added empty / place-holder Pydantic models (classes) for `colormappers`, `filename-formatters`, and `output-formatters`. These are needed for the newly added geoips expand command during the re-validation of the filed input especially after an override.

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,108 @@
...
+class AlgorithmArgumentsModel(PermissiveFrozenModel):
+    """Algorithm step argument definition.
+
+    Pydantic model defining and validating Algorithm step arguments.
+    """
+
+    variables: Optional[List[str]] = Field(
+        [None],
+        description="List of input variables used in algorithm processing",
+    )
```
jsolbrig: I may be forgetting a discussion here but should this be a required field? If not, should it either be an empty list or just `None` rather than `[None]`? I'd lean towards `None` here where, `if not variables:` then operate on all variables.

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,108 @@
...
+    output_data_range: tuple[StrictFloat, StrictFloat] | tuple[None, None] = Field(
+        [None, None],
+        description="list of min and max value for output data product."
+        " This is applied LAST after all other corrections/adjustments."
+        " If None, use data.min() and data.max()",
+    )
```
jsolbrig: Likewise, I think this should default to `None`, not a tuple of `None`s. Maybe I'm forgetting a discussion on this particular variable, though.

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,108 @@
...
+        " * If True, returned data will be inverted"
+        " * If False, returned data will not be inverted",
+    )
+    # This should default to (?, 1000) or (1000, ?) or (None, None)
+    # The upper bound in 'windbarbs_dmw_low.yaml' is of type float
+    pressure_level_range: tuple[float, float] | tuple[None, None] = Field(
+        [None, None],
+        description="list of min and max pressure levels to filter derived motion wind"
+        " retrievals. Defaults to None, which results in using all wind retrievals.",
+    )
```
jsolbrig: I'd also like this to default to `None` rather than `[None, None]`.

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,108 @@
...
+        None,
+        description="Units of input data, for applying "
+        "necessary conversions. Defaults to None, resulting in no unit conversions.",
+    )
+    output_units: str = Field(
+        None,
+        description="Units of input data, for applying "
+        "necessary conversions. Defaults to None, resulting in no unit conversions.",
+    )
+    gamma_list: Optional[List[StrictFloat]] = Field([])
```
jsolbrig: Again, `None` would likely be better here. Setting it to `[]` implies that everything accepts this.

geoips/pydantic_models/v1/colormappers.py
jsolbrig, line 1: Note to @mindyls, these were added as empty for now to support work that others are doing and will have actual arguments added to them shortly.

geoips/pydantic_models/v1/algorithms.py
srikanth-kumar: Setting the default value to `None` would be easiest to move forward. I would prefer updating (quick one and would not take much time)

`if output_data_range == None:`  to `if output_data_range == [None, None]:`. This maintains the data structure consistency. 

Flexible either way, will make changes after your response to this comment

jsolbrig: If we want to switch to `[None , None]`, I would want to modify the behavior of algorithms to allow `[None, 100]` or `[3, None]`. Right now the keyword is all-or-nothing. You either need to provide a full tuple or None. Moving to `[None, None]` makes that non-obvious.

That said, I could see this as an alternative to what I commented on earlier when I suggested that we should eventually deprecate `output_data_range` in favor of `output_minval` and `output_maxval`.

In order to avoid additional code changes for now, though, I'd prefer to have this default to `None`.

srikanth-kumar: Yes, that would trigger more changes; thanks, for pointing that out, Jeremy. I have set the default value to `None`. 

That said, I could see this as an alternative to what I commented on earlier when I suggested that we should eventually deprecate `output_data_range` in favor of `output_minval` and `output_maxval`.  :+1:

geoips/pydantic_models/v1/algorithms.py
srikanth-kumar: Done.

geoips/pydantic_models/v1/algorithms.py
srikanth-kumar: Done

geoips/pydantic_models/v1/algorithms.py
srikanth-kumar: Optional. Updated to just `None`

geoips/pydantic_models/v1/algorithms.py
```diff
@@ -0,0 +1,108 @@
...
+    satellite_zenith_angle_cutoff: Optional[float] = Field(
+        None,
+        description="Cutoff for masking data where satellite zenith angle exceeds"
+        "threshold. None, no masking",
+    )
+    pressure_key: str | None = Field(None)
+    time_key: str = Field(None)
+    time_fcst: StrictInt = Field(-1)
+    time_dim: StrictInt | None = Field(None, alias="Time_Dimension")
+    grid_geo: StrictBool = Field(False)
```
srikanth-kumar: @jsolbrig, I have not set this to `null` as `grid_geo` always defaults to `False`.

srikanth-kumar: Ignore this

srikanth-kumar:
@jsolbrig I have addressed all of your comments. Ready for very quick re-review.

mindyls (commented):
98/98

jsolbrig (approved):

mindyls (approved):

PR #1295, opened 2026-04-16
Replace import of Console with Console from rich.console

biosafetylvl5:
## ✨ Summary

[x] This fixes broken imports on all docs builds..... the old syntax was deprecated in 2020!!
[x] Also fixes memusg import problems by using relative imports

This should pass CI now.


jsolbrig (approved):
I added to this to make the CI start passing. All I needed to do is make some imports in memusg relative.

cpcamacho (approved):

PR #1297, opened 2026-04-20
GeoIPS OBP CLI Updates

evrose54:
## ✨ Summary

Updates to v1 workflow models and 'geoips run order_based <workflow_name>' /
'geoips test workflow <workflow_name>' CLI commands. Updates include refactoring
the 'test' section of a workflow and adding support for overrides in workflow
plugins. These can be specified in the test section of a workflow or at the command
line via '-[Ss]', '-[Kk]', '-[Gg]' flags via the 'geoips run order_based' command.
In order to apply overrides, GeoIPS will fully expand a workflow plugin, replace
relevant portions of the workflow with the contents specified as an override, then
revalidate the modified workflow.

Added documentation and fixed failing unit tests as well. New unit tests to come for
new functionality added.

```
  files:
    added:
      - docs/dev/running_order_based_procflow.rst
      - tests/unit_tests/plugins/yaml/workflows/test_workflow_overrides.py
    modified:
      - docs/source/concepts/functionality/command-line/index.rst
      - geoips/commandline/ancillary_info/cmd_instructions.yaml
      - geoips/commandline/geoips_run.py
      - geoips/commandline/geoips_test.py
      - geoips/interfaces/class_based_plugin.py
      - geoips/interfaces/yaml_based/workflows.py
      - geoips/plugins/modules/procflows/order_based.py
      - geoips/plugins/yaml/workflows/unit_tests/test_bad_test_section.yaml
      - geoips/plugins/yaml/workflows/unit_tests/test_product.yaml
      - geoips/pydantic_models/v1/output_checkers.py
      - geoips/pydantic_models/v1/workflows.py
      - tests/unit_tests/commandline/test_geoips_run.py
      - tests/unit_tests/commandline/test_geoips_test_workflow.py
      - tests/unit_tests/pydantic_models/v1/output_checkers/test_cases_bad.yaml
      - tests/unit_tests/pydantic_models/v1/utils.py
```
---

---
## 🔗 Related Issues

- **Fixes:** `NRLMMD-GEOIPS/geoips#NNN`  
  

---

## 🧪 Testing Instructions

Run ``pytest -v --no-fail-on-missing-data $GEOIPS_PACKAGES_DIR/geoips``.


geoips/plugins/yaml/workflows/unit_tests/test_product.yaml
```diff
@@ -9,8 +9,23 @@ docstring: |
 description: 11.2 µm Infrared.
 test:
   fnames: !ENV ${GEOIPS_TESTDATA_DIR}/test_data_abi/data/goes16_20200918_1950/*
-  command_line_args:
-    compare_path: !ENV ${GEOIPS_PACKAGES_DIR}/geoips/tests/outputs/abi.static.<product>.imagery_clean
+  outputs:
+    abi.Infrared:
+      compare_path: !ENV ${GEOIPS_PACKAGES_DIR}/geoips/tests/outputs/abi.static.Infrared.imagery_clean/20200918.195020.goes-16.abi.Infrared.test_goes16_eqc_3km_day_20200918T1950Z.100p00.noaa.3p0.png
+      token: FAKE_TOKEN
```
evrose54, line 15: output_checker name (potentially with arguments) would be provided.

evrose54: step_id refers to the id of that produced the output that we want to compare against

jsolbrig: I think we might do something like this but I feel like this part is a bit "fuzzy" as you are saying in our meeting @evrose54:

```
test:
  outputs:
    abi.Infrared:  # This is the step ID from the step whose output should be validated
      output_checker:  # If not provided, use a default output_checker
        name: my_output_checker
        arguments: {}
      compare_path: ...
      token: FAKE_TOKEN
```

I don't think we'd use output checkers inside workflow spec sections very often, only test sections. I could see where it might be useful at some point and it would still work, but I think the above would the most common way to use an output checker.

jsolbrig:
```
test:
  outputs:
    abi.Infrared:  # This is the step ID from the step whose output should be validated
      checker_name: my_checker  # defaults to imagery output checker
      arg1: ...
      arg2: ...
      compare_path: ...
      token: FAKE_TOKEN
```

geoips/commandline/geoips_run.py
```diff
@@ -64,6 +66,7 @@ def __call__(self, args):
         args: Namespace()
             - The argument namespace to parse through
         """
+        environ["ORDER_BASED_CALLED"] = "False"
```
jsolbrig: I know we talked about this but adding a comment to document the conversation.

I'd prefer to avoid using the environment to store run-time information. Let's, instead, update all plugins to accept an `order_based` argument. It should default to `False` and get passed to all plugin calls by the OBP.

geoips/commandline/geoips_run.py
```diff
@@ -145,9 +149,162 @@ class GeoipsRunOrderBased(GeoipsExecutableCommand):
...
+        """
+        path = Path(value)
+
+        if not path.exists():
+            raise self.parser.error(f"Input filepath not found: {value}")
+
+        if path.suffix.lower() not in {".json", ".yaml", ".yml"}:
+            raise self.parser.error(
+                f"File must have extension .json, .yaml, or .yml, not {path.suffix}"
+            )
```
jsolbrig: Minor suggestion. It would be good to think about efficiency in cases like this to make sure that the faster checks are performed first. I know that this is a SUPER minor example, but other contexts, this could be important.

The two tests here should be flipped for efficiency and to follow "fail fast". They should look like this:

```
        if path.suffix.lower() not in {".json", ".yaml", ".yml"}:
            raise self.parser.error(
                f"File must have extension .json, .yaml, or .yml, not {path.suffix}"
            )

        if not path.exists():
            raise self.parser.error(f"Input filepath not found: {value}")
```

Existence is a much more expensive test than checking the extension.

geoips/commandline/geoips_run.py
```diff
@@ -169,43 +326,57 @@ def add_arguments(self):
...
             "-f",
             "--filepath",
             default=None,
-            type=json_or_yaml_path,
+            type=self.json_or_yaml_path,
             help=(
                 "A absolute path to the workflow file to operate on."
                 " Cannot be supplied alongside the --workflow or --generated argument."
             ),
         )
```
jsolbrig: How about we combine the `-w` and `-f` options. Let's add a single positional argument before the fnames argument that accepts either the name of a registered workflow or the path to a file containing a workflow. So, the call signature would look like this: `geoips run order_based <workflow> <filenames> [optional arguments]`.

geoips/commandline/geoips_run.py
```diff
@@ -169,43 +326,57 @@ def add_arguments(self):
...
+            "--global-overrides",
+            default=None,
+            nargs="+",
+            type=Dict[str, Any],
+            help=(
+                "One or more global override strings to apply to your workflow. An "
+                "override string should take on the following format:\n "
+                "'<global_variable_name>=<some_value>'"
+            ),
+        )
```
jsolbrig, line 368: Let's add lower-case options for this that can be specified any number of times. Let the upper-case versions accept dictionaries. Let the lower-case versions accept dot-notation to do individual overrides. I want to keep both versions.

In practice, this could look like this for doing two single overrides:

```bash
geoips run order_based my_workflow -s read_abi_netcdf.self_register=LOW -s add_colormap.cmap_name=greys_r
```

Which would be much less difficult to understand and type than:

```bash
geoips run order_based my_workflow -S "{read_abi_netcdf: {self_register: \"LOW\"}, add_colormap: {cmap_name: \"greys_r\"}}"
```

geoips/commandline/geoips_run.py
```diff
@@ -169,43 +326,57 @@ def add_arguments(self):
...
                 "A absolute path to the workflow file to operate on."
                 " Cannot be supplied alongside the --workflow or --generated argument."
             ),
         )
+        # Override arguments
+        self.parser.add_argument(
+            "-S",
+            "--step-overrides",
+            default=None,
+            nargs="+",
```
jsolbrig: I think `nargs` can be removed here. Since you're expecting a dictionary, you just need to accept a single argument.

geoips/commandline/geoips_run.py
jsolbrig: I think we discussed doing this in a follow-on PR which is fine.

geoips/commandline/geoips_run.py
```diff
@@ -230,7 +402,9 @@ def __call__(self, args):
                 )
             # Validate the generated workflow with is_registered set to false as this
             # plugin has been dynamically generated
-            WorkflowPluginModel(**workflow, is_registered=False)
+            WorkflowPluginModel(
+                **workflow, is_registered=False, context={"expand": True}
+            )
```
jsolbrig: Just because this is odd, can you add a comment to explain how `context` works here?

geoips/commandline/geoips_run.py
jsolbrig: Does this break the `--generated` functionality? What does that do and how is it used? I'm a little unclear on it... Is that an instance where we are literally passing a dictionary to the CLI that composes the workflow?

geoips/commandline/geoips_run.py
evrose54: I agree that it's a good habit to put the less expensive things first. Implemented!

geoips/commandline/geoips_run.py
evrose54: ``--generated`` you told me to put together when I created the ``geoips run order_based`` command. You said there might be instances where this command is ran with a workflow that was generated at the time of calling the script (NRT processing most likely). It likely still wouldn't break it. I could combine all three and detect whether the input was a workflow name, file path, or literal dictionary.

geoips/commandline/geoips_run.py
evrose54: You're correct. This was an artifact of the previous implementation. I'll remove that

geoips/commandline/geoips_run.py
evrose54: Can do!

geoips/commandline/geoips_run.py
evrose54: Implemented this. One thing of note; the format  (``-g `` is the same) will look like:

```
-s step:step2:step3:...argument=value
-k kind:argument=value
```

since some step IDs can include a ``.``. Such as ``abi.Infrared``. Colons are the only difference. Also it sort of makes more sense intuitively since they are accessing a nested dictionary.

geoips/pydantic_models/v1/workflows.py
```diff
@@ -628,26 +631,202 @@ def expand_steps(cls, data: dict, info: ValidationInfo):
...
+        final_paths = []
+
+        # cspell:ignore ipath, jpath, jpaths
+        for ipath in filepaths:
+            jpaths = sorted(glob(ipath))
+            for jpath in jpaths:
+                # Don't do pathlib.Path as readers don't expect that.
+                final_paths.append(jpath)
+
+        return final_paths
```
evrose54, line 868: This and the ``fnames`` field in the readers module need to be consolidated... and to use the readers implementation.

geoips/interfaces/class_based_plugin.py
```diff
@@ -160,12 +161,24 @@ def _invoke(self, data=None, *args, **kwargs):
...
+        if environ["ORDER_BASED_CALLED"] == "True":
+            provided_args = set(kwargs)
+            accepted_args = set(list(inspect.signature(self.call).parameters.keys()))
+            unaccepted_args = provided_args - accepted_args
+            for arg in unaccepted_args:
+                provided_args.remove(arg)
+
+            new_kwargs = {kwarg: kwargs[kwarg] for kwarg in provided_args}
+        else:
+            new_kwargs = kwargs
```
evrose54: @jsolbrig can you provide an example of how you'd like to detect if 'order_based' is being ran? I know you don't want to use environment variables but I can't quite remember how you envisioned it being implemented.

jsolbrig: I was thinking that every plugin would accept a keyword like `_obp` that defaults to `False`. Then, the OBP would just pass `_obp=True` when calling plugins. This would get implemented on the base class-based plugin I think.

geoips/commandline/geoips_run.py
```diff
@@ -145,67 +151,391 @@ class GeoipsRunOrderBased(GeoipsExecutableCommand):
...
+        parts = lhs.split(":")
+
+        if len(parts) != 2:
+            raise self.parser.error(
+                f"Invalid key '{lhs}'. Must be in the format of "
+                "'<kind>:<argument_name>'"
+            )
+
+        kind = parts[0]
+        argument = parts[1]
```
jsolbrig, line 228: Can the argument have multiple levels to accommodate nested workflows or products? If "yes", can we update the docstring to indicate the format of nested arguments? If "no", can we update to allow multiple levels?

I'd like the levels to be separated by `.`.

geoips/commandline/geoips_run.py
```diff
@@ -145,67 +151,391 @@ class GeoipsRunOrderBased(GeoipsExecutableCommand):
...
+            # doing a yaml.safe_load attempts to cast the value into its correct type
+            "value": yaml.safe_load(rhs),
+        }
+
+    def step_override_type(self, value: str):
+        """Ensure an override string fits the following format.
+
+        Expected Format
+        ---------------
+        '<step_id>:<string1>:<optional_string2>:<optional_string3>:...=<some_value>'
```
jsolbrig: is there a reason to use `:` rather than `.` here? If not, I'd prefer to use `.`.

evrose54: Yes, the reason is that a step id can have ``.`` in its key. An easy example of this is any product referenced in a workflow, which results in the ``expanded`` step id being ``source_name.product_name``. This is consistent with how we describe a product via the CLI. I.e. ``geoips describe product abi.Infrared``.

geoips/commandline/geoips_run.py
evrose54: So this is for a kind override. In any instance we encounter a step of type ``kind``, we will override or add the specified argument. This includes nested workflows and products. Any instance ``kind`` is encountered, override or add. If you specifically want to override the argument of a specific step (nested or not), you'll have to use the step override format. Does that make sense?

geoips/commandline/geoips_run.py
evrose54: I've combined all of these btw.

geoips/interfaces/class_based_plugin.py
evrose54: I plan to implement this today. Waiting for the integration tests to finish running on #1342 first.

evrose54: This environment variable has been removed and I implemented something akin to what you mentioned above.

geoips/commandline/geoips_run.py
evrose54: Fixed it in this PR

geoips/pydantic_models/v1/workflows.py
evrose54: Consolidated this

geoips/pydantic_models/v1/workflows.py
```diff
@@ -655,26 +659,235 @@ def expand_steps(cls, data: dict, info: ValidationInfo):
...
+
+
+class WorkflowTestModel(FrozenModel):
+    """Model for the test section of GeoIPS workflow plugins."""
+
+    # Not pathlib.Path objects as readers only expect a list of strings
+    fnames: List[str] = Field(
+        ...,
+        description="A list of one or more filepaths to the data used for this test.",
+    )
```
srikanth-kumar, line 780: Are we missing `compare_path` here?

geoips/pydantic_models/v1/workflows.py
evrose54: This readers ``_validate_and_normalize_fnames`` was moved to ``geoips.pydantic_models.v1.bases`` and imported via the readers module. I tried this in the workflows module as well but we require slightly different formats for fnames at the workflow level versus readers level. Unit tests were failing when I tried to use this for both. Therefore, I've kept the code above as it keeps the unit tests passing. We can consolidate at a later point if desired.

geoips/pydantic_models/v1/workflows.py
evrose54: No, that can be supplied in either of the output checker override models. ``OutputCheckerOverride:arguments`` or ``StepOutputOverride``.

docs/dev/running_order_based_procflow.rst
jsolbrig, line 1: This is fine for now. I think that before release of 2.0 we should update this to just be "Running GeoIPS" and deprecate single-source. We could link to single-source here under a section called "Deprecated Methods for Running GeoIPS" or something like that.

Here is an issue for this: https://github.com/NRLMMD-GEOIPS/geoips/issues/1347

geoips/commandline/geoips_run.py
```diff
@@ -145,67 +148,391 @@ class GeoipsRunOrderBased(GeoipsExecutableCommand):
...
+        """Ensure the value provided is a valid pathlib.Path json or yaml file.
+
+        Parameters
+        ----------
+        value: str
+            - The input value for the filepath to typecheck against.
+        no_raise: bool, optional
+            - Whether or not to raise errors if the input value does not match the
+              accepted format. Defaults to False. When True, return False or Path if
+              the input value matches the expected format.
```
jsolbrig: Why does this default to `False`? That seems like unexpected behavior. I would expect this to default to `True` and need to be overridden to not fail.

geoips/commandline/geoips_run.py
```diff
@@ -145,67 +148,391 @@ class GeoipsRunOrderBased(GeoipsExecutableCommand):
         "This warning will be removed once this command is stable.\n"
     )
 
+    def json_or_yaml_path(self, value: str, no_raise=False) -> Path:
```
jsolbrig: How about renaming to `ensure_valid_json_or_yaml_path`? I know it's long but the current name is not easily understandable in the context where it is used.

geoips/commandline/geoips_run.py
```diff
@@ -145,67 +148,391 @@ class GeoipsRunOrderBased(GeoipsExecutableCommand):
...
+        """Ensure the value provided is a valid pathlib.Path json or yaml file.
```
jsolbrig: Let's update this to "Ensure ``value`` is a valid path to a json/yaml file and convert to a Path object."

geoips/commandline/geoips_run.py
jsolbrig: Oh, I misunderstood this. I think that points to a different problem. How about we change this argument from `no_raise` to `warn_on_error` to avoid the double negative, then add a warning when we've decided not to raise an error.

geoips/commandline/geoips_run.py
```diff
@@ -145,67 +148,391 @@ class GeoipsRunOrderBased(GeoipsExecutableCommand):
...
+                is_registered=False,
+                # Adding context in pydantic is akin to passing in values that are
+                # usually None to an Objects __init__ function. It will construct
+                # differently if those parameters are provided. In this case, we are
+                # telling pydantic to expand the workflow, rather than validate just
+                # what's in the data provided
+                context={"expand": True},
+            ).model_dump()
+        # unregistered workflow @ filepath
+        elif self.json_or_yaml_path(value, no_raise=True):
```
jsolbrig: Why `no_raise` here?

geoips/commandline/geoips_run.py
jsolbrig: How hard would it be to switch to using `:`, `-`, or `_` rather than `.` when combining `source_name` and `product_name` so that we can use `.` when doing overrides? I feel like combining `source_name` and `product_name` is a hidden implementation detail while overrides are out-front and visible. I think people would expect overrides to use `.`.

As far as which symbol to use when combining `source_name` and `product_name`, I'm leaning towards `:`.

geoips/commandline/geoips_run.py
```diff
@@ -145,67 +148,391 @@ class GeoipsRunOrderBased(GeoipsExecutableCommand):
...
-        self.obp_mutex_args = self.parser.add_mutually_exclusive_group(required=True)
-        self.obp_mutex_args.add_argument(
-            "-w",
-            "--workflow",
-            default=None,
-            type=str,
+        # Required arguments
+        self.parser.add_argument(
+            "workflow",
+            type=self.workflow_type,
```
jsolbrig, line 285: The help message should be updated here to indicate what can be passed. It can be the name of a registered workflow, the path to an unregistered workflow, or a full dictionary describing the workflow.

This should also remove reference to `--filepath` and `--generated`.

geoips/commandline/geoips_run.py
jsolbrig: Oh, I see. This returns `False` if it can't load something. Maybe the logic here would be easier if we first tried treating an input string as a plugin, then treated it as a path. Just flip the two sections to try loading a plugin first? That would let you get rid of `no_raise` altogether and ignore a lot of my other comments.

geoips/commandline/geoips_run.py
```diff
@@ -145,67 +148,391 @@ class GeoipsRunOrderBased(GeoipsExecutableCommand):
...
+            self.parser.error(
+                "Error: positional argument 'workflow' could not be associated with an"
+                f" accepted type. Input = {value} ; accepted types = "
+                "[str, List[Path], dict]"
+            )
+
+        return workflow
+
+    def dict_type(self, value):
+        """Ensure an dictionary-based override can be cast as a dictionary.
```
jsolbrig, line 151: Please add just a bit more to this docstring to indicate that this is used for validating the input to the `-G`, `-K`, and `-S` options.

geoips/commandline/geoips_run.py
jsolbrig: It does. This can be a future feature. I wonder if we'll want to be able to override `kind` only for specific nested workflows. So, if there are multiple nested workflows, maybe we only want to override the arguments to Interpolator plugins in a specific nested workflow.

Like I said, though, I think this can be a feature that we consider for the future.

You can resolve this after you read this comment.

geoips/commandline/geoips_run.py
evrose54: That seems like a reasonable solution to me!

geoips/interfaces/class_based_plugin.py
```diff
@@ -117,18 +117,29 @@ def _check_name_attribute(cls):
 
     @abstractmethod
     # def call(self, *args: P.args, **kwargs: P.kwargs) -> R:
-    def call(self, *args, **kwargs):
-        """Callable method to be implemented by the plugin class."""
+    def call(self, _obp_initiated=False, *args, **kwargs):
```
jsolbrig: I think knowledge of whether this was called by the OBP shouldn't be required here. If it is, that should require a refactor of the plugin itself. The other three locations are perfect!

geoips/interfaces/class_based_plugin.py
```diff
@@ -152,20 +166,47 @@ def _post_call(self, data=None, *args, **kwargs):
...
+        if _obp_initiated:
+            provided_args = set(kwargs)
+            accepted_args = set(list(inspect.signature(self.call).parameters.keys()))
+            unaccepted_args = provided_args - accepted_args
+            for arg in unaccepted_args:
+                provided_args.remove(arg)
+
+            new_kwargs = {kwarg: kwargs[kwarg] for kwarg in provided_args}
+        else:
+            new_kwargs = kwargs
```
jsolbrig, line 195: I'd like to keep `_invoke()` as clean as possible It should just be the if/else below and nothing else. This logic (and any future logic) should move inside `_pre_call`. We'll talk more about this this afternoon.

evrose54: Yeah, we'll need to talk about this more. I'm not sure it's as simple as just copy pasting this code to the ``_pre_call`` function.

geoips/pydantic_models/v1/output_checkers.py
```diff
@@ -14,19 +14,24 @@
...
 
 from geoips.pydantic_models.v1.bases import FrozenModel
 
 
 class OutputCheckerArgumentsModel(FrozenModel):
     """Output Checker spec (specification) format."""
 
-    compare_path: Optional[FilePath] = Field(
-        strict=True,
+    model_config = ConfigDict(extra="allow")
```
jsolbrig, line 25: This is probably needed for now, but maybe we can get rid of this later after we figure out how to better handle extra override arguments.

geoips/commandline/geoips_run.py
evrose54: Renaming this.

geoips/commandline/geoips_run.py
evrose54: Implemented this

geoips/commandline/geoips_run.py
evrose54: Implemented that. I will say, that the ``else`` statement will never be hit with that logic. I'll need to get your thoughts on what to do there.

geoips/commandline/geoips_run.py
evrose54: Updated the help string and removed those flag references.

geoips/commandline/geoips_run.py
evrose54: Added more context.

geoips/interfaces/class_based_plugin.py
evrose54: Removed this from the call method.

geoips/commandline/geoips_run.py
jsolbrig: This still needs to be addressed

geoips/commandline/geoips_run.py
evrose54: It's now ``source_name:product_name`` and the ``.`` denotes a new step/key

geoips/commandline/geoips_run.py
jsolbrig: Updating. This really should switch these from `nargs="+"` to `action="extend"`.

geoips/plugins/yaml/workflows/unit_tests/test_product.yaml
jsolbrig: Evan is addressing this in a different PR. Closing this for now.

geoips/interfaces/class_based_plugin.py
evrose54: Addressing in a follow on PR.

geoips/commandline/geoips_run.py
evrose54: No longer relevant.

geoips/commandline/geoips_run.py
evrose54: Read. Resolving!

geoips/commandline/geoips_run.py
evrose54: ``action='append'`` was used. It's working now

geoips/commandline/geoips_run.py
jsolbrig: I'm going to leave this open so @mindy can see it when she reviews.

@mindyls this means that references to products will switch from being `<source_name>.<product_name>` to `<source_name>:<product_name>`. I think this is unlikely to impact you but wanted to make sure you're aware in case it does impact you in some way.

geoips/commandline/geoips_run.py
jsolbrig: We spoke about this and @evrose54 has implemented a solution. Resolving.

geoips/interfaces/class_based_plugin.py
jsolbrig: Thanks!

geoips/interfaces/class_based_plugin.py
jsolbrig: The solution here is going to be to move this functionality out of the plugin classes altogether and into Pluginify. These functions should only modify data and should expect the arguments to be pre-filtered. This will happen in Pluginify along with validation of Plugin call signatures against the Pydantic models to ensure that defaults match.

geoips/commandline/geoips_run.py
```diff
@@ -145,67 +147,286 @@ class GeoipsRunOrderBased(GeoipsExecutableCommand):
...
+            type=self.workflow_type,
+            help=(
+                "Workflow instance. Can be the name of a registered workflow plugin, "
+                "a .json or .yaml path to an unregistered workflow plugin, or a "
+                "dictionary that will be literally evaluated as a workflow."
+            ),
+        )
+        self.parser.add_argument(
+            "filenames",
+            nargs="*",
```
jsolbrig: Should this be `nargs=+` rather than `nargs=*`? Presumably we always need to input at least one file.

geoips/interfaces/class_based_plugin.py
evrose54: See https://github.com/NRLMMD-GEOIPS/geoips/issues/1349

geoips/commandline/geoips_run.py
evrose54: Yeah, you're probably right. I can't think of an instance where a reader doesn't expect at least one file path. I'll update that.

evrose54: Addressed.

jsolbrig: Great, thanks @evrose54

jsolbrig (dismissed):

docs/dev/order_based_procflow.rst
```diff
@@ -124,15 +124,24 @@ Output Formatter step in the code block below includes two additional plugins,
...
-          - abi_Infrared.spec.steps.algorithm.output_units='Kelvin'
-        kinds:
-          - readers.self_register=False
-        globals:
-          - sector_list='global_cylindrical'
-          - logging_level='info'
+      outputs:
+        abi:Infrared:
+          compare_path: !ENV ${GEOIPS_PACKAGES_DIR}/geoips/tests/outputs/abi.static.Infrared.imagery_clean/20200918.195020.goes-16.abi.Infrared.test_goes16_eqc_3km_day_20200918T1950Z.100p00.noaa.3p0.png
+          token: FAKE_TOKEN
```
mindyls, line 130: What is the token for?

mindyls (commented):
7/×5

docs/dev/running_order_based_procflow.rst
```diff
@@ -0,0 +1,152 @@
...
+do not need to respecify a flag before each individual override. Just group your
+overrides by type and specify them in a row, then add another flag with overrides if
+required.
+
+Below we'll demonstrate how to use overrides via the GeoIPS CLI.
+
+.. code-block:: bash
+
+  geoips run order_based test_product $GEOIPS_TESTDATA_DIR/test_data_abi/data/goes16_20200918_1950/* \
+    -s abi:Infrared.spec.steps.algorithm.output_units=Kelvin \
```
mindyls, line 64: Note we currently support specifying variables in data sets via VARIABLENAME:DATASETNAME, Eg if 2 different resolution data sets have the same variable name,  you can request the variable from a specific dataset using the : notation.  Do you expect this will cause any issues with the overrides?

docs/dev/order_based_procflow.rst
evrose54: This currently is just a placeholder. It will be used when we have support for tokenizing the output of each step. Quicker than pixel by pixel comparison, and if the tokens don't match, that's when we'd do the more thorough comparison. Again, this is just a placeholder until we have that implemented.

docs/dev/running_order_based_procflow.rst
evrose54: Can you show me somewhere of an example of that? I don't foresee that being an issue, but it could be. We might need to add a ``r'<variable_name>:<dataset_name>'`` in that instance.

mindyls: Remind me next week and I'll find one

mindyls: You can resolve if you remember to remind me 😁

geoips/commandline/geoips_run.py
mindyls: My only concern is related to referencing variables within a dataset by DATASETNAME:VARIABLE NAME (E.G. like viirs maybe that could have the same variable but at different resolutions ? This has been required in the past to request a specific resolution for a variable to ensure we were using what we expected).

This also brings up the question of whether we actually WANT to support the same variable name in different datasets within a datatree / dictionary of datasets ?  And if we do,  do we want to allow specifying DATASETNAME:VARIABLENAME, or handle disambiguation some other way

mindyls: And Also,  would any of the steps etc possibly contain .?

geoips/interfaces/class_based_plugin.py
```diff
@@ -152,20 +159,47 @@ def _post_call(self, data=None, *args, **kwargs):
         return data
 
     # def _invoke(self, data: R, *args: P.args, **kwargs: P.kwargs) -> R:
-    def _invoke(self, data=None, *args, **kwargs):
+    def _invoke(self, data=None, _obp_initiated=False, *args, **kwargs):
+        """Call the main plugin method.
+
+        Additionally, filter out unaccepted arguments if initiated via the OBP.
```
mindyls, line 165: Why do we only filter out arguments from obp?

mindyls: And,  I'm the end it will not filter out arguments at all,  per Jeremycomment?

geoips/plugins/yaml/workflows/unit_tests/test_product.yaml
```diff
@@ -9,8 +9,23 @@ docstring: |
 description: 11.2 µm Infrared.
 test:
   fnames: !ENV ${GEOIPS_TESTDATA_DIR}/test_data_abi/data/goes16_20200918_1950/*
-  command_line_args:
-    compare_path: !ENV ${GEOIPS_PACKAGES_DIR}/geoips/tests/outputs/abi.static.<product>.imagery_clean
+  outputs:
+    abi:Infrared:
```
mindyls, line 13: This works with a colon in the key..

pyproject.toml
```diff
@@ -59,7 +59,7 @@ build-backend = "poetry_dynamic_versioning.backend"                          # o
 # NOTE this setuptools dependency should be removed altogether once meson fortran builds are working
 setuptools = "<70"     # Required for any fortran builds until meson working
 python = ">=3.11.0,<3.13.0"    # mandatory to declare the required python version
-pluginify = { version = "0.1.0.post11.dev0", allow-prereleases = true }       # Required for plugin management and loading
+pluginify = { version = "1.19.1a0", allow-prereleases = true }       # Required for plugin management and loading
```
mindyls, line 62: Are we always going to be specifying a very explicit version of pluginify?

geoips/plugins/yaml/workflows/unit_tests/test_product.yaml
```diff
@@ -23,3 +38,13 @@ spec:
       kind: product
       name: [abi, Infrared]
       arguments: {}
+    # TODO: Implement logic that automatically generates 'output_products'. Right now
+    # if this is not commented out the CI will fail since the 'output_product' paths
+    # differ for CI than on my local machine
```
mindyls, line 43: Is this because the env var based paths are expanded and they don't match between different runs?

Will we still be able to rely on tests to pass if we specify GEOIPS OUTDIRS based output paths ? I would like to retain the ability to ensure paths are produced correctly from integration tests,  and not rely entirely on auto generate output paths for integration tests

docs/dev/running_order_based_procflow.rst
evrose54: Can do! Resolving

geoips/interfaces/class_based_plugin.py
evrose54: It may, but at a different location eventually. The reason we only filter arguments for obi is that global arguments will automatically be passed to all plugins as an argument. That doesn't happen via single_source or config_based due to how they've been coded (all of the conditionals and extra logic and such).

geoips/plugins/yaml/workflows/unit_tests/test_product.yaml
evrose54: That's expected, but I wouldn't be against changing the format eventually, to an underscore or dash. Probably underscore so they'd be valid Python Identifiers. Just going with a colon right now as that's what Jeremy asked for.

pyproject.toml
evrose54: No, we just need to get a stable release up. I need help with that as it always fails for me for some reason

mindyls (commented):
25/25

geoips/plugins/yaml/workflows/unit_tests/test_product.yaml
evrose54: I think it's just due to the fact that the output file path didn't exist on CI. The compare paths do since they're inside of the GeoIPS package, but the output files failed here since they're in GEOIPS_OUTDIRS which I believe is outside the scope of this package's CI.

mindyls: But the CI *will* have GEOIPS OUTDIRS in the end right?

evrose54: I think so? I believe it already does, just that the file didn't exist. @biosafetylvl5 let me know if you disagree or have any more input.

geoips/plugins/yaml/workflows/unit_tests/test_product.yaml
mindyls: _ or - problematic because those can both be in the source or product name ....

geoips/interfaces/class_based_plugin.py
mindyls: Single source also already strips out unsupported kwargs , that is probably why you don't have to do it here

geoips/plugins/yaml/workflows/unit_tests/test_product.yaml
biosafetylvl5: Yes I think we just don't set it in the current (like on `main`) CI doctest because we aren't running any integration tests atm.

mindyls: Confirmed with @biosafetylvl5 this is just temporary, GEOIPS OUTDIRS not currently set on CI SINCE we aren't running integration tests. Will be added

mindyls: Resolve

geoips/commandline/geoips_run.py
jsolbrig: This is mostly to support `Product` and `ProductDefault` plugins. They combine `source_name` and `product_name` to create the plugin name. 

Regarding the same variable in different datasets, I think we do need to support it. For example, if we have IR imagery from both ABI and VIIRS, we would need to support having a hierarchy that looks like this:
- abi
  - my_sector
    - Infrared
- viirs
  - my_sector
    - Infrared
    
or:
- my_sector
  - abi
    - Infrared
  - viirs
    - Infrared

I'm trying to move us towards a more standard `.` notation for getting to those variables. So, you could access the ABI infrared data via `my_sector.abi.Infrared` or something like that. Using `:` for that operation isn't very pythonic.

geoips/interfaces/class_based_plugin.py
jsolbrig: Yep, exactly. This is implementing similar functionality for OBP as single-source uses. I'm hoping that we can move away from the need for this in the future with some careful changes, but we'll see.

geoips/plugins/yaml/workflows/unit_tests/test_product.yaml
jsolbrig: I think the colon will be best here. Keep in mind that this will only get used when we are using a `Product` plugin which will eventually be deprecated and removed. So, colons aren't around forever!

jsolbrig (approved):

geoips/commandline/geoips_run.py
mindyls: So we would basically Always access variables via datasetname.variablename? That probably makes sense anyway

mindyls: Jeremy you can resolve as long as the steroids can never contain . And we would never want it to (should be fine,  since . _ and - are already "protected" delimiters that we avoid for file naming)

mindyls (approved):

geoips/interfaces/class_based_plugin.py
```diff
@@ -137,13 +141,16 @@ def _pre_call(self, data=None, *args, **kwargs):
         return data
 
     # def _post_call(self, data: R, *args, **kwargs) -> R:
-    def _post_call(self, data=None, *args, **kwargs):
+    def _post_call(self, data=None, _obp_initiated=False, *args, **kwargs):
```
srikanth-kumar: Shouldn't this call signature be refactored to make `_obp_initiated=False` a keyword-only argument ?

`def _post_call(self, data=None, *args, _obp_initiated=False, **kwargs)`

geoips/interfaces/class_based_plugin.py
```diff
@@ -152,20 +159,47 @@ def _post_call(self, data=None, *args, **kwargs):
         return data
 
     # def _invoke(self, data: R, *args: P.args, **kwargs: P.kwargs) -> R:
-    def _invoke(self, data=None, *args, **kwargs):
+    def _invoke(self, data=None, _obp_initiated=False, *args, **kwargs):
```
srikanth-kumar: Shouldn't this call signature be refactored to make `_obp_initiated=False` a keyword-only argument ?

`def _invoke(self, data=None, *args, _obp_initiated=False, **kwargs)`

geoips/interfaces/class_based_plugin.py
```diff
@@ -123,12 +123,16 @@ def call(self, *args, **kwargs):
 
     # hooks are intentionally loose; document their accepted kwargs
     # def _pre_call(self, data: R, *args, **kwargs) -> R:
-    def _pre_call(self, data=None, *args, **kwargs):
+    def _pre_call(self, data=None, _obp_initiated=False, *args, **kwargs):
```
srikanth-kumar: Shouldn't this call signature be refactored to make `_obp_initiated=False` a keyword-only argument ?

`def _pre_call(self, data=None, *args, _obp_initiated=False, **kwargs)`

srikanth-kumar:
Running OBP workflow is run on the PR branch, the `ReaderArgumentsModel.fnames` field is populated with an  `argparse.Namespace` object rather than a list of file paths, causing the workflow execution to fail.

In  `geoips/commandline/geoips_run.py file`, line number: 447 should be updated from 

`obp(workflow, args.filenames, args)` to `obp(workflow, fnames=args.filenames, command_line_args=args)` to fix the bug associated in the issue #1355

srikanth-kumar:
`geoips run order_based --help` 
usage: To use, type `geoips run order_based <workflow_type> <filenames> **<order_based_args>`.

**Optional**: should the usage statement be updated to `geoips run order_based <workflow_name> <filenames> **<order_based_args>`

geoips/interfaces/class_based_plugin.py
evrose54: You're correct. Good catch. I'll fix this.

geoips/interfaces/class_based_plugin.py
evrose54: You're correct. Good catch. I'll fix this.

geoips/interfaces/class_based_plugin.py
evrose54: You're correct. Good catch. I'll fix this.

srikanth-kumar (approved):

PR #1336, opened 2026-05-26
Add Colormapper Pydantic model

srikanth-kumar:
## ✨ Summary

This pull request adds a Pydantic-based `ColormapperArgumentsModel` implementation for the Order-Based Procflow (OBP).

---

## 🔗 Related Issues

- **Fixes:** `NRLMMD-GEOIPS/geoips#NNN`  
  

---

## 🧪 Testing Instructions

Let us know how to test/verify your changes if you need anything *beyond* running the integration and unit tests.

Keep it simple and clear—like an empty sky! ☀️

---

## 📸 Output 

Please think about including these in documentation where appropriate.

### 🧪 Demonstrate new test scripts operate as expected
- Command line outputs or imagery outputs demonstrating passed tests.
- For imagery outputs commited to the repository, include clean imagery using very small test sectors with minimal formatting (no extra YAML metadata, or extra image annotations!).
- If you've added new image products, remember to create tests in:
  - `<repo>/tests/scripts/<testname>.sh`
  - And update `<repo>/tests/integration/integration_tests.py`

### 🖼️ Pretty imagery demonstrating functionality

- You can drop "pretty" imagery outputs directly in this section - these are for reference only, and can be used to share the beauty of your updates with others! Imagery you drop directly in the PR can have any formatting you wish - the prettier the better.

### 🧙🔎 Image difference files

Sometimes matplotlib or numpy will cause very minor differences in image outputs, causing an updated image with no visible change within the github files changed tab
In these cases, it can sometimes be useful to drop the image diff output from the geoips run directly in this section (lives in tests/outputs/[folder]/diff_*)
These image diff outputs are a faded out version of the image, with bright red highlighting where there were changes in the image between the old and new version.

---

💖🌟🌎🌟💖


geoips/pydantic_models/v1/colormappers.py
```diff
@@ -3,14 +3,89 @@
...
 
 class ColormapperArgumentsModel(PermissiveFrozenModel):
     """Colormapper step argument definition.
 
     Pydantic model defining and validating Colormapper step arguments.
     """
 
-    pass
+    cbar_full_width: bool = Field(
+        default=PLUGIN_PROVIDED,
```
evrose54: Based on existing colormappers, this should default to ``False``

geoips/pydantic_models/v1/colormappers.py
```diff
@@ -3,14 +3,89 @@
...
+        description='"Extend the colorbar across the full width of the image"',
+    )
+    cbar_label: str = Field(
+        default=PLUGIN_PROVIDED,
```
evrose54, line 28: Based on existing colormappers, this should default to ``None``

geoips/pydantic_models/v1/colormappers.py
```diff
@@ -3,14 +3,89 @@
...
+            " can also specify directly within 'set_ticks_kwarg'"
+        ),
+    )
+    cmap_name: str = Field(
+        default=PLUGIN_PROVIDED,
+        description=(
+            "Specify the name of the resulting matplotlib colormap. If no ascii_path"
+            " specified, will use builtin matplotlib colormap of name cmap_name."
+        ),
+    )
```
evrose54, line 62: Not sure how we want to handle this, but I'd probably think of a different default value here. We can't make it required since not all colormappers require this argument, but it would fail for ``matplotlib_linear_norm`` since ``'plugin_provided'`` is not a 'named colormap'. Maybe do a before validator based on the color mapper plugin referenced? And if camp_name is in that plugin's call signature, then validate it accordingly?

geoips/pydantic_models/v1/colormappers.py
```diff
@@ -3,14 +3,89 @@
...
+    data_range: tuple[float, float] | str = Field(
+        default=PLUGIN_PROVIDED,
+        description=(
+            "Min and max value for colormap"
+            "matplotlib.colors.Normalize(vmin=min_val, vmax=max_val)"
+        ),
+    )
+    pressure_range_legend: List[str] | None = Field(
+        None,
+        description="List of strings that are used for setting the cbar tick labels",
```
evrose54, line 81: Is this description correct?

geoips/pydantic_models/v1/colormappers.py
```diff
@@ -3,14 +3,89 @@
...
+    cmap_source: str = Field(default=PLUGIN_PROVIDED)
+
+    colorbar_kwargs: dict | None = Field(
+        None, description="keyword arguments to pass through directly to 'fig.colorbar'"
+    )
+    create_colorbar: bool | None = Field(
+        True, description="Specify whether the image should contain a colorbar or not."
+    )
+    data_range: tuple[float, float] | str = Field(
+        default=PLUGIN_PROVIDED,
```
evrose54, line 73: I'd probably find a different default value for this. 'plugin_provided' isn't a tuple to begin with. Maybe just ``None``?

geoips/pydantic_models/v1/colormappers.py
srikanth-kumar: I had to opt for PLUGIN_PROVIDED because this argument has different default values in these two plugins

- `winds/dmw_wind_levels.py` has `cbar_full_width: True,`
- `matplotlib_log_norm.py` has `cbar_full_width=False`,

geoips/pydantic_models/v1/colormappers.py
srikanth-kumar: Opted for "PLUGIN_PROVIDED" as `cbar_label` takes different values across different plugins.

geoips/pydantic_models/v1/colormappers.py
srikanth-kumar: To support 'plugin_provided', I have already added `str` as an additional accepted type , which should address the concern you raised.

geoips/pydantic_models/v1/colormappers.py
srikanth-kumar: Yes

geoips/pydantic_models/v1/colormappers.py
jsolbrig: @mindyls Can you see whether you agree with me that this should just default to `False` for all colormappers and require an override to be `True` for a specific application?

jsolbrig: I'm fine with leaving this for now as long as we make an issue to revisit this one later.

srikanth-kumar: Thank you, @evrose54, for catching this. @jsolbrig I've looked at this again and I agree with you that the default value should be `False`.

I've set the default value to False. Ready to be resolved.

geoips/pydantic_models/v1/colormappers.py
srikanth-kumar: I've set the default value to False. Ready to be resolved.

srikanth-kumar: Please ignore the previous comment, this still needs the default value as PLUGIN_PROVIDED.

- `matplotlib_log_norm.py` : `cbar_label=None,`
- `cmap_37H_Legacy.py` : `cbar_label="TB (K)"`

geoips/pydantic_models/v1/colormappers.py
srikanth-kumar: Related to #1340 

This above discusses the problem and has been addressed in the PR #1345.

To summarize: 

- Any key with value "PLUGIN_PROVIDED" in the workflow dict will be removed before OBP invocation.  
- This secondary default value should be as the default value for the arguments with different default values across different plugins of the same interface type.
- Once the default value for a given argument becomes consistently same across plugins, `PLUGIN_PROVIDED` can be replaced with the corresponding default value.

geoips/pydantic_models/v1/colormappers.py
evrose54: Gotcha. That works for me.

geoips/pydantic_models/v1/colormappers.py
evrose54: Resolving!

geoips/pydantic_models/v1/colormappers.py
evrose54: Works for me. Thanks Kumar. Resolving.

geoips/pydantic_models/v1/colormappers.py
evrose54: Works for me. Thanks Kumar. Resolving.

geoips/pydantic_models/v1/colormappers.py
```diff
@@ -3,14 +3,89 @@
...
+# Python Standard Libraries
+from typing import List
+
+# Third-Party Libraries
+from pydantic import Field
+
 # GeoIPS imports
 from geoips.pydantic_models.v1.bases import PermissiveFrozenModel
 
+PLUGIN_PROVIDED = "plugin_provided"
```
evrose54: Should this point at #1345 and use the constant you added there?

evrose54 (dismissed):
Approved. Thanks for doing this @srikanth-kumar. I left one more comment that might be a good idea to address.

geoips/pydantic_models/v1/colormappers.py
srikanth-kumar: I've imported the constant from the constants module.

evrose54 (dismissed):

jsolbrig (approved):

geoips/pydantic_models/v1/colormappers.py
```diff
@@ -13,4 +20,71 @@ class ColormapperArgumentsModel(PermissiveFrozenModel):
     Pydantic model defining and validating Colormapper step arguments.
     """
 
-    pass
+    cbar_full_width: bool = Field(
+        default=False,
```
mindyls, line 24: Will there be some way for us to fail if an individual plugin attempts to set a default value that does not match the default value in the pydantic model?  Because i can see this causing confusion for all of the pydantic models... this does not need to be addressed right now,  but we probably need to create an issue and address it at some point.

mindyls (approved):

geoips/pydantic_models/v1/colormappers.py
jsolbrig: Yes, we do have plans for that. We'll do it in a different PR, but the intent is to add something that would validate plugin call signatures at both registration-time and load-time.

Issue #1340, opened 2026-05-28
Add `PLUGIN_PROIVDED` as an additional default value in OBP Pydantic models.


srikanth-kumar:
### Description

Currently, we are using `None` as the default value in two different cases: 

- When `None` is the actual default value for a plugin argument across all plugins of the same type.
   -  `area_def=None` across all reader plugins
- When there are different default values for the same argument across different plugins
  - For instance, observe the `data_range` in both the call signatures below 
    - def call(data_range=[0, 8])
    -  def call(data_range=[125, 310], cbar_label="TB (K)"):

Note: The other default value cases such as False across for a given argument across all the plugins of the same type is outside the scope of this issue and need not be handled. 

A Pydantic field must be either:

- required : user must set an explicit value
- optional : a default value is provided

Pydantic does not provide a built-in method / way for users to leave the field blank / "unset" while keeping the field optional. As a result, the Pydantic model must assign some default value, and currently `None` is being used for that purpose. 

Python would treat `None` as a meaningful value for the given keyword argument and would not override `None` it with the default value defined in the call signature. 

From the above examples ( in the second case): `None`  will not allow the plugin call signature to fall back to its own default value, such as `data_range=[0, 8]`. 

### Code to demonstrate issue

```
from typing import Any, List
from pydantic import BaseModel, Field

class ReaderArgumentsModel(BaseModel):
    """Reader step argument definition.

    Pydantic model defining and validating Reader step arguments.
    """

    variables: List[str] = Field(
        None,
        description="List of variables to read",
    )
    metadata_only: bool = Field(False, description="Read metadata only.")
    self_register: str = Field(None, description="Enable self-registration.")

workflow = ReaderArgumentsModel(variables=["B14BT"], metadata_only=True)

# workflow at this point : {variables=['B14BT'] metadata_only=True self_register=None}

def test_fucntion(variables, metadata_only, self_register="registered"):
    "Simple test call"
    print(self_register)

result =  test_fucntion(**workflow.model_dump())
# value of self_register in the workflow at this point
# self_resgiter = None
```

### Solution: 

Python treats `None` as an explicitly provided keyword argument value rather than as an omitted argument. 

We propose a new default value called `PLUGIN_PROVIDED` that would help us to sanitize the workflow: drop the fields/arguments with the value = "PLUGIN_PROVIDED".  This will make these fields omitted letting the call signature fill in the required default value. 

### Code to demonstrate solution 

```
[...]

PR #1345, opened 2026-06-01
Add `PLUGIN_PROIVDED` as an additional default value in OBP Pydantic models.

srikanth-kumar:
## ✨ Summary

This PR introduces a new default value called `PLUGIN_PROVIDED` for use in OBP Pydantic models. 

`PLUGIN_PROVIDED`  is intended for workflow arguments whose actual default values must be defined by the underlying plugin call signatures and may across plugins of the same interface type. During the workflow sanitization, fields with value "PLUGIN_PROVIDED" are removed from the workflow before invocation. This causes the corresponding keyword arguments to be omitted entirely, allowing the plugin call signature to populate the appropriate default value. 

This change addresses a limitation in the current Pydantic workflow generation approach. Pydantic does not provide a built-in mechanism for leaving an optional field truly "unset" while still keeping the field optional within the model definition. As a result, optional fields currently require an explicit default value, with None being used in many cases.

However, Python treats None as an explicitly provided keyword argument value. Consequently, when None is passed to a plugin call, the default value defined in the function signature is not applied.

By introducing PLUGIN_PROVIDED and sanitizing those fields before plugin invocation, the workflow can preserve plugin-defined defaults while still maintaining valid optional Pydantic fields within the OBP models

In addition, this PR also introduces, a designated file at `geoips/constants.py` to define constants used across GeoIPS for both scientific and data-processing purposes. This is to ensure that the value of constant stays consistently constant throughout the codebase. 

- [ ✅] **Tests**: All tests and CI pass (e.g., full, base, extra, etc.)
    - [ ✅] **Full test**: Yes, full test *is passing*.
- [ ✅] **Integration Tests**: Integration tests added/updated for new/modified functionality *OR* no new functionality
- [ ✅] **Other Repos**: I have updated other repositories to handle this change *OR* my package doesn't impact other plugin packages.

---

## 🔗 Related Issues

- **Fixes:** NRLMMD-GEOIPS/geoips#1340  
  

---

## 🧪 Testing Instructions

Run `pytest` at `/geoips/tests/unit_tests/plugins/modules/procflows`

image

---

## 📸 Output 

Please think about including these in documentation where appropriate.

### 🧪 Demonstrate new test scripts operate as expected
- Command line outputs or imagery outputs demonstrating passed tests.
- For imagery outputs commited to the repository, include clean imagery using very small test sectors with minimal formatting (no extra YAML metadata, or extra image annotations!).
- If you've added new image products, remember to create tests in:
  - `<repo>/tests/scripts/<testname>.sh`
  - And update `<repo>/tests/integration/integration_tests.py`

### 🖼️ Pretty imagery demonstrating functionality

- You can drop "pretty" imagery outputs directly in this section - these are for reference only, and can be used to share the beauty of your updates with others! Imagery you drop directly in the PR can have any formatting you wish - the prettier the better.

### 🧙🔎 Image difference files

Sometimes matplotlib or numpy will cause very minor differences in image outputs, causing an updated image with no visible change within the github files changed tab
In these cases, it can sometimes be useful to drop the image diff output from the geoips run directly in this section (lives in tests/outputs/[folder]/diff_*)
These image diff outputs are a faded out version of the image, with bright red highlighting where there were changes in the image between the old and new version.

---

💖🌟🌎🌟💖

### DIFF
--- a/docs/source/releases/latest/1345-add-plugin-provided-as-secondary-default-value-in-obp-pydantic-models.yaml
+++ b/docs/source/releases/latest/1345-add-plugin-provided-as-secondary-default-value-in-obp-pydantic-models.yaml
@@ -1,0 +1,21 @@
+feature:
+- title: 'Add PLUGIN_PROIVDED as an additional default value in OBP Pydantic models'
+  description: 'This PR introduces a new default value called PLUGIN_PROVIDED that would help us to sanitize the workflow: drop the fields/arguments with the value = "PLUGIN_PROVIDED". This will make these fields omitted letting the call signature fill in the required default value.'
+  files:
+    deleted:
+    - ''
+    moved:
+    - ''
+    added:
+    - 'tests/unit_tests/plugins/modules/procflows/conftest.py'
+    - 'tests/unit_tests/plugins/modules/procflows/test_obp_utils.py'
+    - 'geoips/docs/source/releases/latest/update-dev-staging.yaml'
+    - 'geoips/plugins/modules/procflows/obp_utils.py'
+    modified:
+    - 'geoips/plugins/modules/procflows/order_based.py'
+  related-issue:
+    number: null
+    repo_url: ''
+  date:
+    start: null
+    finish: null
--- a/docs/source/releases/latest/1345-add-plugin-provided-as-secondary-default-value-in-obp-pydantic-models.yaml
+++ b/docs/source/releases/latest/1345-add-plugin-provided-as-secondary-default-value-in-obp-pydantic-models.yaml
@@ -1,6 +1,15 @@
 feature:
 - title: 'Add PLUGIN_PROIVDED as an additional default value in OBP Pydantic models'
-  description: 'This PR introduces a new default value called PLUGIN_PROVIDED that would help us to sanitize the workflow: drop the fields/arguments with the value = "PLUGIN_PROVIDED". This will make these fields omitted letting the call signature fill in the required default value.'
+  description: | 
+    'This PR introduces a new default value called PLUGIN_PROVIDED that would help us to
+     sanitize the workflow: drop the fields/arguments with the value = "PLUGIN_PROVIDED".
+      This will make these fields omitted letting the call signature fill in the required
+       default value.'
+
+    Note-1: This is a breaking change.
+    Note-2: If we do not specify the default value as "PLUGIN_PROVIDED" and set it to
+      `None`, Pydantic-Python behavior will not allow the plugin-defined defaults to
+      actually take effect.
   files:
     deleted:
     - ''
--- a/docs/source/releases/latest/1345-add-plugin-provided-as-secondary-default-value-in-obp-pydantic-models.yaml
+++ b/docs/source/releases/latest/1345-add-plugin-provided-as-secondary-default-value-in-obp-pydantic-models.yaml
@@ -2,14 +2,15 @@ feature:
 - title: 'Add PLUGIN_PROIVDED as an additional default value in OBP Pydantic models'
   description: | 
     'This PR introduces a new default value called PLUGIN_PROVIDED that would help us to
-     sanitize the workflow: drop the fields/arguments with the value = "PLUGIN_PROVIDED".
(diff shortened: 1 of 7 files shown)


docs/source/releases/latest/1345-add-plugin-provided-as-secondary-default-value-in-obp-pydantic-models.yaml
```diff
@@ -0,0 +1,22 @@
+feature:
+- title: 'Add PLUGIN_PROIVDED as an additional default value in OBP Pydantic models'
+  description: 'This PR introduces a new default value called PLUGIN_PROVIDED that would help us to sanitize the workflow: drop the fields/arguments with the value = "PLUGIN_PROVIDED". This will make these fields omitted letting the call signature fill in the required default value.'
```
mindyls: Maybe leave a note in the release note that if we don't do this, pydantic will secretly replace the plugin's kwarg defaults with it's own defaults.

geoips/plugins/modules/procflows/obp_utils.py
```diff
@@ -0,0 +1,80 @@
...
+def remove_keys_with_default_value_plugin_provided(workflow_dict: dict) -> dict:
+    """Sanitize the workflow by removing keys with value 'PLUGIN_PROVIDED'.
+
+    'PLUGIN_PROVIDED' acts as a secondary default value in the OBP. It is used when the
+    same argument across different plugins of the same type has different default
+    values.
+
+    This function recursively removes workflow arguments whose value is
+     'PLUGIN_PROVIDED'. The plugin call signature would process these arguemnts as
+     ommited and populates the default value accordingly.
```
mindyls: Maybe explicitly leave a note in here that if we don't do this, pydantic will secretly replace the plugin's kwarg default values with its own.

mindyls (dismissed):

geoips/plugins/modules/procflows/obp_utils.py
evrose54: Looks like these are indented improperly. Also, ``arguments``, not ``arguemnts``

geoips/plugins/modules/procflows/obp_utils.py
```diff
@@ -0,0 +1,80 @@
...
+    Parameters
+    ----------
+    workflow: dict
+        The workflow plugin dictionary to process.
+
+    Returns
+    -------
+    dict
+        workflow dictionary with all keys containing the value 'plugin_provided'
+         removed.
```
evrose54: Same here; weird indentation

evrose54 (dismissed):
Approved, but see comments. Minor changes needed. Good work!

geoips/plugins/modules/procflows/obp_utils.py
srikanth-kumar: Note-1: This is a breaking change.
    Note-2: If we do not specify the default value as "PLUGIN_PROVIDED" and set it to
      `None`, Pydantic-Python behavior will not allow the plugin-defined defaults to
      actually take effect.
      
Left explicit notes as suggested.  Good recommendation; thanks, @mindyls

docs/source/releases/latest/1345-add-plugin-provided-as-secondary-default-value-in-obp-pydantic-models.yaml
srikanth-kumar: Done

geoips/plugins/modules/procflows/obp_utils.py
srikanth-kumar: Fixed. Thanks, Evan for catching the indentation and the typo.

geoips/plugins/modules/procflows/obp_utils.py
srikanth-kumar: Fixed.

evrose54 (approved):

mindyls (approved):

jsolbrig (approved):

PR #1413, opened 2026-07-09
Workflow Argument Overrides

evrose54:
## ✨ Summary

Adds support for argument overrides to steps in an embedded workflow. Additionally,
added logic to handle 'depends_on' fields for embedded workflows. This way you
can specify depends on to the outer workflow, and the first data step of the
embedded workflow must add a depends_on field using [_input] * len(outer_dependencies)
to make sure data is passed appropriately to the first data step.

```
  files:
    modified:
      - geoips/interfaces/class_based/workflow.py
      - geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
      - geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
      - geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
      - geoips/plugins/yaml/workflows/unit_tests/Infrared/VIIRS-Infrared.yaml
      - geoips/plugins/yaml/workflows/unit_tests/Infrared/output-imagery-annotated.yaml
      - geoips/pydantic_models/v1/workflows.py
```

---

---
## 🧪 Testing Instructions

Run ``pytest -v $GEOIPS_PACKAGES_DIR/geoips --no-fail-on-missing-data``
---

## 📸 Output 

``geoips test workflow ABI-Infrared``

goes_abi_infrared_full_run

``geoips test workflow AHI-Infrared``
himawari_ahi_infrared_full_run

``geoips test workflow VIIRS-Infrared``
jpss_viirs_infrared_full_run


geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
```diff
@@ -0,0 +1,32 @@
...
+docstring: |
+  A workflow stub for producing a scaled infrared product and its colormap.
+spec:
+  steps:
+    interpolator:
+      kind: interpolator
+      name: interp_nearest
+      depends_on: [_input, _input]
+      arguments:
+        varlist: ['B14BT']
```
mindyls, line 15: Maybe this should be 'SENSOR_SPECIFIC' or something instead of 'B14BT' to make it clear it needs to be overridden on a sensor-by-sensor basis?

geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
```diff
@@ -0,0 +1,32 @@
...
+    apply_single_channel:
+      kind: algorithm
+      name: single_channel
+      arguments:
+        output_data_range: [-90.0, 30.0]
```
mindyls, line 20: I think it makes sense to have a default here (and it is typically unchanged, and makes sense for it to be consistent)

geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
mindyls: Basically just force people to override the variable name?

geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
```diff
@@ -0,0 +1,40 @@
...
+      depends_on: []
+      arguments: {}
+    infarared_workflow:
+      kind: workflow
+      name: Infrared-Base
+      depends_on: [reader, sector]
+      arguments:
+        interpolator.varlist: ["B13BT"]
+        apply_single_channel.output_data_range: [-90, 30]
+        infrared_colormap.data_range: [-90, 30]
```
mindyls, line 29: These don't strictly have to be overridden correct?  Only the varlist ?

geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
```diff
@@ -0,0 +1,40 @@
...
+    reader:
+      kind: reader
+      name: ahi_hsd
+      arguments:
+        chans: ["B13BT"]
+    sector:
+      kind: sector
+      name: japan
+      depends_on: []
+      arguments: {}
```
mindyls: Are you able to do a list of sectors here and produce multiple outputs?  And does this actually sector the data, or just attach an area definition to the datatree?

geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
evrose54: Yup!

geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
evrose54: We definitely can. These are here just to demonstrate that you can do overrides like this. Not required by any means

geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
evrose54: I can do that. @jsolbrig thoughts on that?

geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
evrose54: Unfortunately I think you'd need to specify multiple sector steps to grab multiple area definitions. Then do interpolation and whatever other steps are required to get it to the new sector grid. Then multiple output steps. This is definitely something we'll want to clean up in the future post 2.0 alpha. It doesn't actually sector the data, just attaches an area def to the tree.

geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
mindyls: Agreed and understood, I was just making sure I understood properly.

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: geoips/v1
+kind: Workflow
+# In SSP can specify product without specifying source, then rely on the SSP to find the
```
srikanth-kumar, line 3: I agree with the comment, but is there a reason it's placed at this location in the file? Or is it intended to briefly explain the hyphenated combination of `ABI` and `Infrared` in the mapping `name`?

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
```diff
@@ -0,0 +1,40 @@
...
+      kind: reader
+      name: abi_netcdf
+      arguments:
+        chans: ['B14BT']
+    sector:
+      kind: sector
+      name: goes_east
+      depends_on: []
+      arguments: {}
+    infarared_workflow:
```
srikanth-kumar: infrared

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
```diff
@@ -0,0 +1,40 @@
...
+  steps:
+    reader:
+      kind: reader
+      name: abi_netcdf
+      arguments:
+        chans: ['B14BT']
+    sector:
+      kind: sector
+      name: goes_east
+      depends_on: []
```
srikanth-kumar, line 22: We should remove `depends_on` and `arguments` here.

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
```diff
@@ -0,0 +1,40 @@
...
+      arguments:
+        interpolator.varlist: ["B14BT"]
+        apply_single_channel.output_data_range: [-90, 30]
+        infrared_colormap.data_range: [-90, 30]
+    apply_imagery_annotated:
+      kind: workflow
+      name: output-imagery-annotated
+      depends_on: [apply_single_channel, gridlines, features, infrared_colormap, sector]
+      arguments:
+        apply_imagery_annotated:
```
srikanth-kumar, line 37: Is this validated by `OutputFormatterArgumentsModel`? If so, how does the validation model handle the additional / nested level (`apply)imagery_annotated`) of nesting?

geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
```diff
@@ -0,0 +1,40 @@
...
+  steps:
+    reader:
+      kind: reader
+      name: ahi_hsd
+      arguments:
+        chans: ["B13BT"]
+    sector:
+      kind: sector
+      name: japan
+      depends_on: []
```
srikanth-kumar: [same](https://github.com/NRLMMD-GEOIPS/geoips/pull/1413/changes#r3577400706)

geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
```diff
@@ -0,0 +1,40 @@
...
+      arguments: {}
+    infarared_workflow:
```
srikanth-kumar: [same](https://github.com/NRLMMD-GEOIPS/geoips/pull/1413/changes#r3577390476)

geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
```diff
@@ -0,0 +1,32 @@
...
+family: order_based
+interface: workflows
+docstring: |
+  A workflow stub for producing a scaled infrared product and its colormap.
+spec:
+  steps:
+    interpolator:
+      kind: interpolator
+      name: interp_nearest
+      depends_on: [_input, _input]
```
srikanth-kumar, line 13: What is the purpose of `[_input, _input]` here?

geoips/plugins/yaml/workflows/unit_tests/Infrared/output-imagery-annotated.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: geoips/v1
+kind: Workflow
+name: output-imagery-annotated
+family: order_based
+interface: workflows
+docstring: |
+  A workflow stub for outputting data using the imagery annotated output formatter.
```
srikanth-kumar, line 7: How is this workflow stub intended to be used? Is it meant to be referenced by other workflows as a nested workflow, such as a product sub-workflow?

geoips/plugins/yaml/workflows/unit_tests/Infrared/VIIRS-Infrared.yaml
```diff
@@ -0,0 +1,40 @@
...
+      arguments:
+        interpolator.varlist: ["I05BT"]
+        apply_single_channel.output_data_range: [-90, 30]
+        infrared_colormap.data_range: [-90, 30]
+    apply_imagery_annotated:
+      kind: workflow
+      name: output-imagery-annotated
+      depends_on: [apply_single_channel, gridlines, features, infrared_colormap, sector]
+      arguments:
+        apply_imagery_annotated:
```
srikanth-kumar, line 35: **Lines 32 and 37:** Why do we have the same mapping keys at both levels of nesting?

srikanth-kumar:
Nice work, Evan! I've left a few comments for your review.

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
evrose54: Huh, I didn't write that comment. I don't know how that got there.

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
evrose54: Good catch, I'll fix that

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
evrose54: Can do

geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
evrose54: Can do

geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
evrose54: Will fix

geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
evrose54: An interpolator step requires two inputs (data, sector) in order to run. Validation would fail if depends on was empty or not exactly length 2. Sense this is more of a workflow stub (like a product default), than a fully defined workflow, I added these in order to pass validation at runtime. ``_input`` are used as the depends on steps as those are reserved keywords in the workflow pydantic code which denotes that the dependency doesn't have to be present in the current workflow, since they're out of scope.

geoips/plugins/yaml/workflows/unit_tests/Infrared/VIIRS-Infrared.yaml
evrose54: I just named them the same as that is what the resulting step id will be. The embedded workflow is loaded in place and has its arguments overridden accordingly.

geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
jsolbrig: I'm tempted to just leave it out entirely, but I can see how "SENSOR_SPECIFIC" might be more clear.

geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
jsolbrig: At the moment, you would need to run the workflow several times, once per sector. I plan to fix that. For now, the quick an dirty fix will be to have the OBP itself loop over the sectors and call this workflow multiple times. When we implement Split operators, one of them will explicitly split over sector_list.

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
jsolbrig: FYI, there is an explicit difference between `depends_on: []` and not setting it at all. If it isn't set at all, it will implicitly depend on the step before it. An empty list says "I depend on nothing!"

geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
jsolbrig: `depends_on: []` here says "I don't depend on anything" while removing it says "I depend on the step before me". I'm not sure that I like this behavior, but it is what it is at the moment.

That said, I think `sector` should come before `reader` most of the time because it reduces what the reader is required to do. The reader actually uses the sector. At least some of them do!

geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
jsolbrig: Is it required to use `_input` twice or could it just be specified once here?

geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
christopher-selman: Whatever we do here gonna have to change it in 1380 I think

geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
christopher-selman: Since we're in a pinch here to get this done, I'm okay with this for now-- just make sure an issue is opened

jsolbrig: Created #1448

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
jsolbrig: Removed.

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
jsolbrig: Fixed

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
jsolbrig: This actually passes the arguments specified such that they override arguments to the step of the same name in the embedded workflow. That said, I'm not sure that I like the syntax here. I'll create an issue to come back and revisit this. Thankfully, I'd call this a minor hidden issue that, while important, is not important for 2.0-alpha.

jsolbrig: Created #1449

geoips/plugins/yaml/workflows/unit_tests/Infrared/AHI-Infrared.yaml
jsolbrig: Fixed

geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
jsolbrig: Created #1450

geoips/plugins/yaml/workflows/unit_tests/Infrared/output-imagery-annotated.yaml
jsolbrig: Yes, exactly right.

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
```diff
@@ -0,0 +1,40 @@
...
+      name: Infrared-Base
+      depends_on: [reader, sector]
+      arguments:
+        interpolator.varlist: ["B14BT"]
+        apply_single_channel.output_data_range: [-90, 30]
+        infrared_colormap.data_range: [-90, 30]
+    apply_imagery_annotated:
+      kind: workflow
+      name: output-imagery-annotated
+      depends_on: [apply_single_channel, gridlines, features, infrared_colormap, sector]
```
srikanth-kumar: `geoips.errors.PluginResolutionError: step 'apply_imagery_annotated' depends_on 'apply_single_channel': 'apply_single_channel' is not a step in the referenced workflow; valid steps: ['apply_imagery_annotated', 'infarared_workflow', 'reader', 'sector']`

srikanth-kumar: geoips.errors.PluginResolutionError: step 'apply_imagery_annotated' depends_on 'infrared_workflow.apply_single_channel': container step 'infrared_workflow' has no inline spec to resolve nested steps

srikanth-kumar: I'm applying the necessary fixes to get the CI tests passing, but these changes do not implement the underlying functionality.

jsolbrig:
After discussion, I think we need to update our format a bit. The current format is still confusing. We should still expand at call-time so that workflows are static, though.

Here's a suggested format:

```python
apiVersion: geoips/v1
interface: workflows
family: order_based
name: mwr_tb325_15_imagery_clean
docstring: |
  AWS MWR 325.15 GHz brightness temperature orbital imagery over global cylindrical sector.
package: geoips
test:
  fnames: !ENV ${GEOIPS_TESTDATA_DIR}/test_data_arctic_weather_satellite/data/20250416/W_NO-KSAT-Tromso%2CSAT%2CAWS1-MWR-1B-RAD_C_OHB__20250416161751_G_O_20250416142833_20250416160447*.nc
  outputs:
    output_formatter:
      full_test_policy: on_token_mismatch # always | never
      compare_path: !ENV ${GEOIPS_PACKAGES_DIR}/geoips/tests/outputs/aws.orbital.imagery_clean
      output_checker_name: image
  arguments:
    global:
      product_name: test_TB325_1
      reader_defined_area_def: True
    my_workflow.read_abi:
      my_argument: "test_override"
arguments:
  global:
    product_name: TB325_1
    reader_defined_area_def: False
  my_workflow.read_abi:
    my_argument: "top_level_override"
spec:
  steps:
    ...
```

Some notes:
- Arguments from the CLI would override anything provided by the `arguments` section when calling `geoips run`.
- Arguments from the `test` section would, likewise, take precedence over the `arguments` section. Essentially, the `test` section provides the same information for `geoips test` and the CLI arguments do for `geoips run`.
- For simplicity, let's remove the `kind` overrides.

srikanth-kumar:
https://github.com/NRLMMD-GEOIPS/geoips/pull/1413#issuecomment-5146834548

Thanks for putting this together, @jsolbrig. I have few questions / clarifications: 

1.  "For simplicity, let's remove the kind overrides." do you mean removing support for both the `-K` dict and `-k` string formats? 
2.  Is the `arguments.global` section is still expected to support `-g` and `-G` global overrides ?
3.  Are the step override args provided via the CLI expected to be added under  `arguments.global` ? 

Sorry, I misread as if there is no arguments section under test section. 

Can we optimize 


```
test:
  fnames: !ENV ${GEOIPS_TESTDATA_DIR}/test_data_arctic_weather_satellite/data/20250416/W_NO-KSAT-Tromso%2CSAT%2CAWS1-MWR-1B-RAD_C_OHB__20250416161751_G_O_20250416142833_20250416160447*.nc
  outputs:
    output_formatter:
      full_test_policy: on_token_mismatch # always | never
      compare_path: !ENV ${GEOIPS_PACKAGES_DIR}/geoips/tests/outputs/aws.orbital.imagery_clean
      output_checker_name: image
  arguments:
    global:
      product_name: test_TB325_1
      reader_defined_area_def: True
    my_workflow.read_abi:
      my_argument: "test_override"
arguments:
  global:
    product_name: TB325_1
    reader_defined_area_def: False
  my_workflow.read_abi:
    my_argument: "top_level_override"
   ```
   
   to 
   
```
test:
  fnames: !ENV ${GEOIPS_TESTDATA_DIR}/test_data_arctic_weather_satellite/data/20250416/W_NO-KSAT-Tromso%2CSAT%2CAWS1-MWR-1B-RAD_C_OHB__20250416161751_G_O_20250416142833_20250416160447*.nc
  outputs:
    output_formatter:
      full_test_policy: on_token_mismatch # always | never
      compare_path: !ENV ${GEOIPS_PACKAGES_DIR}/geoips/tests/outputs/aws.orbital.imagery_clean
      output_checker_name: image
arguments:
  global:
    product_name: TB325_1
    reader_defined_area_def: False
  my_workflow.read_abi:
    my_argument: "top_level_override"
   ```  
This gives each section a single responsibility and avoids duplication of arguments under two sections:
  - `test` contains only test-specific configuration. 
  - `arguments` contains the global arguments, step-level overrides, and global overrides

evrose54:
@srikanth-kumar I agree that it makes more sense to me to just have one ``arguments`` section. I know the ``test: arguments`` field would take precedence over the outer-level ``arguments`` field, but IMO, this is making things overly complex. It's also confusing to have two sections that can contain the same content. Just as a user I think this is getting too into the weeds. If needed, you can just add CLI flags to override what is present in the original ``arguments`` section. @jsolbrig, is that an ok solution to you?

biosafetylvl5:
@dgshanee posted this related issue #1486

geoips/plugins/yaml/workflows/unit_tests/Infrared/Infrared-Base.yaml
evrose54: I'm honestly not sure off the top of my head. It's been a while since I wrote this. Let me try it out and I'll get back to you.

srikanth-kumar:
Running into embedded override issue as mentioned [here](https://github.com/NRLMMD-GEOIPS/geoips/pull/1475#issuecomment-5590252454).

srikanth-kumar:
`geoips run obp abi_static_infrared_imagery_annotated_enhanced ${GEOIPS_TESTDATA_DIR}/test_data_abi/data/
goes16_20200918_1950/* -s  abi_Infrared.interpolator.arguments.test=test` 

Bug:

```
08_190744      readers.py:158  WARNING: 'chans' is deprecated and will be removed in GeoIPS 2.0. Use 'variables' instead.
08_190744      readers.py:158  WARNING: 'chans' is deprecated and will be removed in GeoIPS 2.0. Use 'variables' instead.
  File "/home/kumar/miniforge3/envs/geoips311/bin/geoips", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/local/home/kumar/geoips/geoips/geoips/commandline/commandline_interface.py", line 270, in main
    args = geoips_cli.execute_command()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local/home/kumar/geoips/geoips/geoips/commandline/commandline_interface.py", line 124, in execute_command
    self.GEOIPS_ARGS.exe_command(self.GEOIPS_ARGS)
  File "/local/home/kumar/geoips/geoips/geoips/commandline/geoips_run.py", line 415, in __call__
    workflow = self._apply_overrides(workflow, args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local/home/kumar/geoips/geoips/geoips/commandline/geoips_run.py", line 398, in _apply_overrides
    WorkflowPluginModel(**workflow, is_registered=False)
  File "/local/home/kumar/miniforge3/envs/geoips311/lib/python3.11/site-packages/pydantic/main.py", line 253, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for WorkflowPluginModel
spec.steps.abi_Infrared.interpolator
  Extra inputs are not permitted [type=extra_forbidden, input_value={'arguments': {'arguments': {'test': 'test'}}}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/extra_forbidden
    ```

**Comment by evrose54** [comment]
@srikanth-kumar:

@biosafetylvl5, @jsolbrig, and myself talked extensively about how the ``depends_on`` field should be specified for complex workflows (I.e. workflows with nested workflows). We've come to a consensus that the format needs to be modified, so for now intend to merge this PR as-is, accepting that nested workflow dependencies **will not** work for the time being. This will be addressed in a future PR once we've solidified the exact format we want to implement for these use cases.

Additionally, I will create an issue requesting that we add a new feature for overriding a step's plugin ``name`` in the future. This way we can override what plugin is used for any given step at the command line. Please resolve comments you see fit regarding what I've said in this message and ping me if you think there are any changes that must be made to this PR prior to merging it.

Here is that issue:

https://github.com/NRLMMD-GEOIPS/geoips/issues/1495

**Review by srikanth-kumar — DISMISSED** [review]

**Comment by biosafetylvl5** [comment]
Evan is fixing the last two failing unit tests. Then this is ready to go!

File: geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
Comment (jsolbrig) [inline geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml:1]: Do these belong in a directory called "unit_tests" or do they belong in "integration_tests/" instead?

File: geoips/pydantic_models/v1/workflows.py [hunk 15]
```diff
@@ -903,6 +904,30 @@ def expand_step(
         else:
             steps = cls.expand_steps(plugin.get("spec"), info)["steps"]
 
+        for key, value in override_args.items():
+            if isinstance(value, dict):
+                # occurs for override arguments formatted like such
+                # step_id:
+                #   argument_name: value
+                for argument_name, argument_value in value.items():
+                    steps[key]["arguments"][argument_name] = argument_value
```
jsolbrig: It would be more pythonic to use duck typing here.

```
# Try the dict form
try:
    for argument_name, argument_value in value.items():
        steps[key]["arguments"][argument_name] = argument_value
except AttributeError:
    # Try the doc-notation form
    try:
        step_id, argument_name = key.split(".")
    except AttributeError, ValueError:
        raise something useful
```

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
evrose54: Sorry, I forgot about this comment. I'll remove that line.

geoips/pydantic_models/v1/workflows.py
jsolbrig: I mainly brought this up because it occurred to me that there is no error handling here in the case where `value` is not a dict and `key` doesn't split into two strings.

jsolbrig: Possible problems in the dot notation form include:
- Dot is missing: would raise `ValueError: not enough values to unpack (expected 2, got 1)`
- More than two dots: would raise `ValueError: too many values to unpack (expected 2)`
- Key is not a string (probably not possible due to pydantic but worth checking anyway: would raise `AttributeError: 'int' object has no attribute 'split'`

geoips/pydantic_models/v1/workflows.py
```diff
@@ -903,6 +904,30 @@ def expand_step(
...
+
+        if _inputs and kind == "workflow":
+            for step_id, step in steps.items():
+                if step.get("kind") not in [
+                    "feature_annotator",
+                    "gridline_annotator",
+                    "colormapper",
+                ]:
+                    steps[step_id]["depends_on"] = _inputs
+                    break
```
jsolbrig, line 942: Please comment here to describe what is being done and why.

geoips/pydantic_models/v1/workflows.py
evrose54: addressed! I agree that duck typing here is more useful.

geoips/pydantic_models/v1/workflows.py
evrose54: Can do.

geoips/plugins/yaml/workflows/unit_tests/Infrared/ABI-Infrared.yaml
evrose54: Either or. I say we leave them where they are now and move them later once everything is finalized.

mindyls (approved):

jsolbrig (approved):

PR #1433, opened 2026-07-13
Remove outputs field from steps

jsolbrig:
## ✨ Summary

The functionality provided by the `outputs` field in workflow steps has been superseeded by expanded use of `depends_on`. This PR removes all remaining references to `outputs`.

Also, all tests should pass on this PR! 🤞


geoips/plugins/yaml/workflows/abi_infrared_test.yaml
```diff
@@ -0,0 +1,84 @@
...
+        data_range: [-90.0, 30.0]
+
+    coverage_checker:
+      kind: coverage_checker
+      name: masked_arrays
+      depends_on: [algorithm, sector]
+      arguments:
+        # I'm not sure if this is the correct behavior here or if this just needs to be
+        # documented as the behavior from single_channel
+        variable_name: data
```
mindyls, line 67: This works?

geoips/plugins/yaml/workflows/abi_infrared_test.yaml
```diff
@@ -0,0 +1,84 @@
...
+      name: abi_netcdf
+      depends_on: []
+      arguments:
+        chans: [B14BT]
+
+    sector:
+      kind: sector
+      name: conus
+      depends_on: []
+      arguments: {}
```
mindyls, line 31: Does this actually sector the data, or just attach an area definition to the datatree?  Is it exactly the same whether you put the sector section before or after the reader?

Is it possible to replicate the single source functionality here of producing multiple sectors in a single run?

geoips/utils/yaml_utils.py
mindyls, line 1: Do we need to retain this?  Should we just be importing from geoips_yaml_utils at this point?

mindyls (dismissed):
36/36

evrose54 (dismissed):
LGTM. Note the failing doc build.

geoips/utils/yaml_utils.py
jsolbrig: Maybe. This was a rush job. I'd say we should merge this as-is and create an issue to assess.

geoips/plugins/yaml/workflows/abi_infrared_test.yaml
jsolbrig: This does work. I can run this workflow end-to-end.

Apparently the behavior in the OBP is to set the algorithm output to `data` if the algorithm results in a single data array.

We probably need to examine this behavior and ensure that it's what we want. I'm not convinced...

geoips/plugins/yaml/workflows/abi_infrared_test.yaml
jsolbrig: It just creates a datatree containing an area definition. It doesn't actually matter where it is placed so long as the plugins that require it specify this step_id in `depends_on`. That said, readers should probably depend on this for sectored reads and the like. I'm working my way towards that today in other PRs.

geoips/utils/yaml_utils.py
coleman-m: I think that this really does need to be removed, along with `tests/unit_tests/utils/test_yaml_utils.py`. And any import sites which call `import geoips.utils.yaml_utils as yaml` should be changed to `import geoips_yaml_utils as yaml`.

geoips/plugins/yaml/workflows/abi_infrared_test.yaml
mindyls: Also, when required, how do you actually tell the OBP to actually "sector" the data, as opposed to just attach the area definition to the data tree?  Would we have to create an algorithm that sectors data, which would return a new data array that is sectored appropriately?  Do we need to turn some of the operations that are internal to the single source or config based procflows into algorithms that can be added as a step to the OBP workflows?

geoips/plugins/yaml/workflows/abi_infrared_test.yaml
mindyls: Maybe specify the output variable name from the upstream steps, and input variable names from downstream steps?  For now it can be "data", but perhaps open an issue to address this after the workshop?

jsolbrig: 100% an issue to solve for 2.0-beta. I'll make an issue for this.

mindyls: Please link to issue here when you create it, then this can be resolved.

mindyls (approved):

geoips/utils/yaml_utils.py
mindyls: Looks like it's done!  Resolve this if its ready to go Coleman.

mindyls (approved):

geoips/plugins/yaml/workflows/abi_infrared_test.yaml
jsolbrig: I'm planning to add this to the readers in a PR later today rather than require a separate plugin. It will go into the _post_call so that we don't end up needing to reinvent the same behavior for every reader.

geoips/plugins/yaml/workflows/abi_infrared_test.yaml
jsolbrig: Created #1435

evrose54 (approved):

PR #1458, opened 2026-07-21
2.0.0a0 - geoips - VERSION RELEASE - OBP, CBP, pluginify, etc!

biosafetylvl5:
- [x] This is the 2.0.0a0 release.


.github/versions/tagged_version
```diff
@@ -1 +1 @@
-1.19.0
+2.0.0a
```
mindyls: This should be 2.0.0a0, though I'm not 100% sure the CI will tag it appropriately either way

docs/update_release_note_index.py
```diff
@@ -137,6 +137,8 @@ def main(index_filename, release_note_path):
         # appropriate alpha version, and remove the aN from the mmb_version_num
         if "a" in version_num:
             mmb_version_num, alpha = version_num.split("a")
+            if alpha == "":
```
mindyls, line 140: Why not just add the 0 to the version number?

mindyls (commented):
I think we should version 2.0.0a0

docs/update_release_note_index.py
biosafetylvl5: 🤷 I like the 2.0.0a syntax but you're right we can just add the 0

.github/versions/tagged_version
biosafetylvl5: Changing!

docs/update_release_note_index.py
biosafetylvl5: I'm leaving this but I added the extra 0

.github/versions/tagged_version
biosafetylvl5: Changed

coleman-m (dismissed):
LGTM

.github/versions/tagged_version
mindyls: Thank you

docs/update_release_note_index.py
mindyls: sounds good

mindyls (dismissed):

mindyls (approved):

jsolbrig (approved):

